PR #1, opened 2021-01-20
TFM: configurable memory layout

oyvindronningstad:
Only the top-level CMakeLists.txt needs to be reviewed.

PR #4, opened 2021-02-01
Support MCUBOOT_IMAGE_NUMBER and CMAKE_BUILD_TYPE

oyvindronningstad:
Additions necessary to work with the modem enabled.

based on #1.


hakonfam (approved):

tejlmand:
Is this required for 1.5 ?

I made a comment here: https://github.com/zephyrproject-rtos/trusted-firmware-m/pull/43/#issuecomment-773560663
that would mean that those changes would be a single commit directly in Zephyr repo.

And that would anyway have to wait until Zephyr 2.5 is out.
But if this is needed for NCS 1.5, then maybe just do those changes as a temp-hack that will simply be dropped on the following upmerge when the CMake code is in Zephyr repo.

CMakeLists.txt
```diff
@@ -206,6 +206,9 @@ if (CONFIG_BUILD_WITH_TFM)
     set(TFM_IMAGE_NUMBER_ARG
         MCUBOOT_IMAGE_NUMBER ${CONFIG_TFM_MCUBOOT_IMAGE_NUMBER})
   endif()
+  if (DEFINED CMAKE_BUILD_TYPE)
+    set(BUILD_TYPE_ARG CMAKE_BUILD_TYPE ${CMAKE_BUILD_TYPE})
```
tejlmand, line 210: See comment here: https://github.com/zephyrproject-rtos/trusted-firmware-m/pull/44#discussion_r570518633

Is there a special NCS use-case and is it needed for NCS 1.5 ?

tejlmand (changes requested):

ioannisg (approved):

CMakeLists.txt
oyvindronningstad: I was using it to get the RAM usage down to fit with the modem RAM. It turned out not to be needed, but might as well add it now that I completed it.

tejlmand (approved):

PR #5, opened 2021-02-10
[nrf fromlist] CMakeLists.txt: Add signing code consolidated from zephyr boards

oyvindronningstad:
Mirror of https://github.com/zephyrproject-rtos/trusted-firmware-m/pull/45

In the process, make a few changes:
- Change some names of files created.
- Minimize the number of files created.
- Use hex files instead of bin files. This is so we don't need to know
  the offset when creating hex files from bin files.

Also add signing for MCUBOOT_IMAGE_NUMBER=1 based on the code from the
v2m_musca_b1 board, though, this board does not build with =1 now
because of (I assume) some flash aliasing which places the S and NS
images 0x10000000 apart, where the manual algorithm places them next to
each other. It builds with =2, though.

Signed-off-by: Øyvind Rønningstad <oyvind.ronningstad@nordicsemi.no>

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -229,4 +229,110 @@ if (CONFIG_BUILD_WITH_TFM)
   )
 
   zephyr_link_libraries(tfm_api)
+
+  # Set default image versions if not defined elsewhere
+  if (NOT DEFINED TFM_IMAGE_VERSION_S)
+    set(TFM_IMAGE_VERSION_S 0.0.0+0)
+  endif()
+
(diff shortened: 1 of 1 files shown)


tejlmand:
with https://github.com/zephyrproject-rtos/trusted-firmware-m/pull/45 being closed without merge, I guess this should be a `[nrf temphack]` instead.

tejlmand (approved):
LGTM, but commit should be `[nrf temphack]` now.

PR #6, opened 2021-02-19
Fix TFM build on SES Windows

oyvindronningstad:


andvib (approved):

ioannisg (approved):

trusted-firmware-m/toolchain_GNUARM.cmake
```diff
@@ -13,8 +13,13 @@ endif()
 
 set(CMAKE_SYSTEM_NAME Generic)
 
-set(CMAKE_C_COMPILER ${CROSS_COMPILE}-gcc)
-set(CMAKE_ASM_COMPILER ${CROSS_COMPILE}-gcc)
+if (${CMAKE_HOST_SYSTEM_NAME} STREQUAL "Windows")
+    set(CMAKE_C_COMPILER ${CROSS_COMPILE}-gcc.exe)
```
tejlmand, line 17: This would be a more correct  way to fix this issue:
```
find_program(CMAKE_C_COMPILER ${CROSS_COMPILE}-gcc)
find_program(CMAKE_ASM_COMPILER ${CROSS_COMPILE}-gcc)
```

then there is no need for testing platform, and it even would allow a caller to fixed the CMAKE_C_COMPILER in the cache, if need be.

tejlmand (changes requested):
The title seems wrong to me.

This also fails on command line in windows, with CMake / ninja.

trusted-firmware-m/toolchain_GNUARM.cmake
oyvindronningstad: The CROSS_COMPILE var also contains the path. Will this still work or do we need to split it @tejlmand?

tejlmand: no, that will work.
It just mean CMake won't look into other places for finding the program.
As example:
```
find_program(TEST /bin/test)
```
will not find candidates under `/usr/bin/test`.

oyvindronningstad: Ok, but that sounds like what we want. Should I add checks for `-NOTFOUND`?

tejlmand: > Ok, but that sounds like what we want. Should I add checks for `-NOTFOUND` @tejlmand?

That is always a good thing to do as it can also detect wrong values being provided by `${CROSS_COMPILE}`, even though it wasn't verified before.

oyvindronningstad:
> The title seems wrong to me.

Can you elaborate @tejlmand?

trusted-firmware-m/toolchain_GNUARM.cmake
oyvindronningstad: Done

tejlmand (approved):

tejlmand:
> > The title seems wrong to me.
> 
> Can you elaborate @tejlmand?

The title says:
> Fix TFM build on SES Windows

But the commits in this PR says:
> CMakeLists.txt: Pass ZEPHYR_BASE to TFM build

has nothing to do with windows.

> trusted-firmware-m: Use find_program for C compiler 

Which is of course related to Windows due to missing `.exe` but the title for the PR seems to hint something wrong related to SES Windows.

PR #7, opened 2021-03-03
kconfig: tfm: support for custom CMake flags when building TF-M

tejlmand:
This commit allows a user or subsystem to specify additional CMake flags
to be given to the TF-M build.

The CMake flags can be provided in two ways:
- TFM_EXTRA_CMAKE_FLAGS_KCONFIG: which allows for additional CMake flags
  as a space separeted list, like: "-DARG0=val0 -DARG1=val1"
- TFM_EXTRA_CMAKE_FLAGS_CMAKE: which allows a CMake subsystem to specify
  additional flags on the `tfm_build_target` using the property
  `CMAKE_FLAGS`.

The Kconfig option allows users to set flags, whereas the CMake options
allows for more complex build integration, such as the usage of
generator expressions.

Signed-off-by: Torsten Rasmussen <Torsten.Rasmussen@nordicsemi.no>

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -18,6 +18,9 @@
 # ISOLATION_LEVEL: The TF-M isolation level to use
 # REGRESSION: Boolean if TF-M build includes building the TF-M regression tests
 # BL2: Boolean if the TF-M build uses MCUboot. Default: True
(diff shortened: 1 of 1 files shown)


ioannisg:
This can probably be closed, since there is a PR in the zephyr side now

PR #10, opened 2021-03-25
[nrf temphack]: Adding configurable BL2 config

frkv:
-Changing from static definition to configurable CMake variable
 called MCUBOOT_MBEDCRYPTO_CONFIG_PATH

Signed-off-by: Frank Audun Kvamtrø <frank.kvamtro@nordicsemi.no>

### DIFF
--- a/trusted-firmware-m/bl2/CMakeLists.txt
+++ b/trusted-firmware-m/bl2/CMakeLists.txt
@@ -60,7 +60,7 @@ target_compile_definitions(bl2_mbedcrypto_config
     INTERFACE
         $<$<STREQUAL:${MCUBOOT_SIGNATURE_TYPE},RSA>:MCUBOOT_SIGN_RSA>
         $<$<STREQUAL:${MCUBOOT_SIGNATURE_TYPE},RSA>:MCUBOOT_SIGN_RSA_LEN=${MCUBOOT_SIGNATURE_KEY_LEN}>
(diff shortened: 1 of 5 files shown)


oyvindronningstad (approved):
Can you open a patch upstream?

tejlmand (approved):

frkv:
> Can you open a patch upstream?

Sorry for delaying this. We have had a chat on this today and have come to the conclusion that having "PATH" instead of "STRING" is difficult for us

We feed TF-M build system with a different location for the mbed TLS (they call it MBEDCRYPTO). If we feed the full PATH then we make assumption on where the build folders are even before the build system has started building the two (or more) variants of the mbed TLS (MBEDCRYPTO). One of the items we discussed is to give the header-files in their own build-folder, but that would decouple it more

We will phrase the PR to showcase the issue so that we combine this with the ability to "replace" mbed TLS (MBEDCRYPTO)

ioannisg:
@frkv what is the status of this PR? Which parts are required for your work in nrfconnect/sdk-nrf#4156? If not everything is required, perhaps we can clean this up and get it properly reviewed and merged? If some parts are not temphacks, could we send patches to the upstream module?

ioannisg:
Rebased that @frkv 
@oyvindronningstad @tejlmand it seems to me this PR can be merged to master independently; it does not break anything, AFAIK.
Btw do we agree that the two commits are "upstreamable"? In that case we can open an upstream patch for these.

oyvindronningstad (approved):

PR #15, opened 2021-05-12
Upmerge 12 v2021

nvlsianpu:


ioannisg:
I think this looks pretty straight forward. I only wonder if it would be better to revert first and then re-apply the temp-hack patches merged in #10

andvib (approved):
Passes CI

oyvindronningstad (approved):

ioannisg (approved):

frkv (approved):
LGTM

ioannisg:
Good to go @nvlsianpu

PR #17, opened 2021-05-20
[nrf fromlist]: Enable CCM as default for protected storage

Vge0rge:
[Upstream PR](https://review.trustedfirmware.org/c/TF-M/trusted-firmware-m/+/10125)

-Enables AES-CCM as the default AEAD algorithm
 for nRF9160 and nRF5340

Ref: NCSDK-9006
Manifest PR: [pull/4645](https://github.com/nrfconnect/sdk-nrf/pull/4645)

Signed-off-by: Georgios Vasilakis <georgios.vasilakis@nordicsemi.no>

### DIFF
--- a/trusted-firmware-m/platform/ext/target/nordic_nrf/common/nrf9160/config.cmake
+++ b/trusted-firmware-m/platform/ext/target/nordic_nrf/common/nrf9160/config.cmake
@@ -9,3 +9,4 @@
 set(SECURE_UART1                        ON         CACHE BOOL      "Enable secure UART1")
 set(PSA_API_TEST_TARGET                 "nrf"      CACHE STRING    "PSA API test target")
 set(ITS_NUM_ASSETS                      "5"        CACHE STRING    "The maximum number of assets to be stored in the Internal Trusted Storage area")
+set(PS_CRYPTO_AEAD_ALG                  PSA_ALG_CCM CACHE STRING    "The AEAD algorithm to use for authenticated encryption in Protected Storage")


trusted-firmware-m/platform/ext/target/nordic_nrf/common/nrf5340/config.cmake
```diff
@@ -9,3 +9,4 @@
 set(SECURE_UART1                        ON         CACHE BOOL      "Enable secure UART1")
 set(PSA_API_TEST_TARGET                 "nrf"      CACHE STRING    "PSA API test target")
 set(ITS_NUM_ASSETS                      "5"        CACHE STRING    "The maximum number of assets to be stored in the Internal Trusted Storage area")
+set(PS_CRYPTO_AEAD_ALG                  PSA_ALG_CCM CACHE STRING    "The AEAD algorithm to use for authenticated encryption in Protected Storage")
```
oyvindronningstad, line 12: Do we need to switch to CCM on 53?

ioannisg: Good question. @Vge0rge ?

Vge0rge: I thought that we will use the same algorithm for both just to have to have the same scheme. We definitely don't "need" to do that for the 5340 since GCM is supported so I will just update the PR.

frkv (approved):
LGTM

ioannisg (approved):
Fine with me. I wonder if this can go upstream, as I do not see anything controversial.
@oyvindronningstad ^^

frkv:
> Fine with me. I wonder if this can go upstream, as I do not see anything controversial.
> @oyvindronningstad ^^

Agreed. Totally generic solution without impact if you don't use it

oyvindronningstad (approved):

oyvindronningstad:
> 
> 
> > Fine with me. I wonder if this can go upstream, as I do not see anything controversial.
> 
> Agreed. Totally generic solution without impact if you don't use it

+1

ioannisg:
@frkv @Vge0rge we are discussing whether we should also force the nRF5340 default configuration, to make sure it is not overwritten by the PROFILE definitions. (By default we do not set up a profile type, but if someone does, then we risk having the wrong setting in 5340)

ioannisg:
https://review.trustedfirmware.org/c/TF-M/trusted-firmware-m/+/10125

ioannisg:
By the way, @Vge0rge this needs a sauce tag as well (I guess "from list")

Vge0rge:
Will be moved to the sdk-nrf repo.

PR #18, opened 2021-05-24
[nrf fromtree] platform: nordic_nrf: add non-secure storage partition

Damian-Nordic:
If the newly added CMake variable NRF_NS_STORAGE=ON, then configure the last 32KiB of the flash to be non-secure.
This is to allow non-secure applications to use the non-volatile storage.

Upstream PR: https://review.trustedfirmware.org/c/TF-M/trusted-firmware-m/+/11039

Signed-off-by: Damian Krolik <damian.krolik@nordicsemi.no>


trusted-firmware-m/platform/ext/target/nordic_nrf/common/nrf5340/target_cfg.c
```diff
@@ -196,6 +196,10 @@ enum tfm_plat_err_t spu_init_cfg(void)
         memory_regions.non_secure_partition_limit);
     spu_regions_sram_config_non_secure(NS_DATA_START, NS_DATA_LIMIT);
 
+#if defined(NS_STORAGE_OFFSET) && defined(NS_STORAGE_LIMIT)
```
oyvindronningstad, line 199: Check whether the size is non-zero as well.

Damian-Nordic: Done

oyvindronningstad (approved):

ioannisg (commented):
Discussed offline with @Damian-Nordic 
The change here makes sense: we need to tell TF-M that a portion of flash at the end of the flash area is dedicated to file system/storage partition, and needs to be given to the application.

However, I'd like to see this work being "upstreamable". This involves, representing the aforementioned area in the memory regions, properly, so a single call to the spu API for flash configuration should suffice.

We might need to introduce new macros in flash_layout.h and/or region_defs.h, specific to nRF platforms, describing the storage region. This can also be done upstream. The macros can be either left empty, or have a default value. We could also introduce a CMAKE variable to enable the feature (e.g. to "use" a storage partition).

Eventually, what we would like to keep downstream is the integration with the PArtition Manager.

oyvindronningstad:
I agree with @ioannisg about upstreaming.

The changes I'd like to see to make it more upstreamable:
- Add to nRF9160 as well.
- As @ioannisg says, add macros to flash_layout.h, the size could be 0 by default. The Cmake variable can definitely be considered, but it's not standard in TFM for partitions to be controlled via Cmake.
- Have NRF_ as a prefix on the macro names.

Damian-Nordic:
@ioannisg @oyvindronningstad I updated this PR and https://github.com/nrfconnect/sdk-nrf/pull/4614. Let me know if that's more or less what you suggested.

trusted-firmware-m/platform/ext/target/nordic_nrf/common/nrf5340/partition/flash_layout.h
```diff
@@ -148,6 +150,11 @@
                                          FLASH_ITS_AREA_SIZE)
 #define FLASH_NV_COUNTERS_AREA_SIZE     (FLASH_AREA_IMAGE_SECTOR_SIZE)
 
+/* Non-secure storage region */
+#define NRF_FLASH_STORAGE_AREA_OFFSET   (FLASH_TOTAL_SIZE - \
+                                         NRF_FLASH_STORAGE_AREA_SIZE)
+#define NRF_FLASH_STORAGE_AREA_SIZE     (0x8000)   /* 32 KB */
```
oyvindronningstad, line 156:
```suggestion
#define NRF_FLASH_NS_STORAGE_AREA_OFFSET   (FLASH_TOTAL_SIZE - \
                                            NRF_FLASH_STORAGE_AREA_SIZE)
#define NRF_FLASH_NS_STORAGE_AREA_SIZE     (0x8000)   /* 32 KB */
```

Damian-Nordic: Done, also added the CMake variables.

trusted-firmware-m/platform/ext/target/nordic_nrf/common/nrf9160/partition/flash_layout.h
```diff
@@ -149,6 +151,11 @@
                                          FLASH_ITS_AREA_SIZE)
 #define FLASH_NV_COUNTERS_AREA_SIZE     (FLASH_AREA_IMAGE_SECTOR_SIZE)
 
+/* Non-secure storage region */
+#define NRF_FLASH_NS_STORAGE_AREA_OFFSET    (FLASH_TOTAL_SIZE - \
```
oyvindronningstad, line 155: This should be calculated from `(FLASH_NV_COUNTERS_AREA_OFFSET + FLASH_NV_COUNTERS_AREA_SIZE)` to avoid overlapping regions. Also, add a check after, to see that the regions don't go over `FLASH_TOTAL_SIZE`.

oyvindronningstad (commented):
Thinking about it, IMO we should try to reduce the size of the patch since it is downstream (for the time being), and anyways NRF-specific. That means removing the changes in region_defs.h and the linker script, and also the entry in the `memory_region_limits`. So basically just using the defines in flash_layout.h directly. @ioannisg please comment, and if there's a reason I've missed to keep the changes please let me know.

trusted-firmware-m/platform/ext/target/nordic_nrf/common/nrf9160/partition/flash_layout.h
Damian-Nordic: Storage partition is always at the end so I would prefer leave it as it is and simply add a check that NV Counters and NS Storage don't overlap. Would that be fine?

trusted-firmware-m/platform/ext/target/nordic_nrf/common/nrf5340/config.cmake
```diff
@@ -9,3 +9,4 @@
 set(SECURE_UART1                        ON         CACHE BOOL      "Enable secure UART1")
 set(PSA_API_TEST_TARGET                 "nrf"      CACHE STRING    "PSA API test target")
 set(ITS_NUM_ASSETS                      "5"        CACHE STRING    "The maximum number of assets to be stored in the Internal Trusted Storage area")
+set(NS_STORAGE                          OFF        CACHE BOOL      "Enable non-secure storage partition")
```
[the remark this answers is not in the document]
Damian-Nordic: ok, will do

Damian-Nordic:
@oyvindronningstad I previously had a solution which configures the NS region directly (although I missed macros in flash_layout.h). Previously I thought the rework was needed to keep things consistent, so now I need help with understanding why the current approach is undesired :)

trusted-firmware-m/platform/ext/target/nordic_nrf/common/nrf9160/partition/flash_layout.h
oyvindronningstad: That works. As long as we prevent overlapping I don't have a strong preference.

oyvindronningstad (approved):
I see that @ioannisg requested this specifically, so I won't interfere. The current solution is more consistent, and I was wondering if it's worth it to sacrifice that for a smaller downstream patch. I don't see it becoming a big issue either way though, so we can keep it. Sorry for the noise.

trusted-firmware-m/platform/ext/target/nordic_nrf/common/nrf5340/config.cmake
```diff
@@ -9,3 +9,4 @@
 set(SECURE_UART1                        ON         CACHE BOOL      "Enable secure UART1")
 set(PSA_API_TEST_TARGET                 "nrf"      CACHE STRING    "PSA API test target")
 set(ITS_NUM_ASSETS                      "5"        CACHE STRING    "The maximum number of assets to be stored in the Internal Trusted Storage area")
+set(NRF_NS_STORAGE                      OFF        CACHE BOOL      "Enable non-secure storage partition")
```
hakonfam, line 12: Does this enable the storage partition? Or does it make an existing storage partition available as non-secure?

trusted-firmware-m/platform/ext/target/nordic_nrf/common/nrf5340/partition/flash_layout.h
```diff
@@ -40,7 +41,8 @@
  * 0x000f_0000 Protected Storage Area (16 KB)
  * 0x000f_4000 Internal Trusted Storage Area (8 KB)
  * 0x000f_6000 NV counters area (4 KB)
- * 0x000f_7000 Unused
+ * 0x000f_7000 Unused (4 KB)
+ * 0x000f_8000 Non-secure storage (32 KB)
```
hakonfam, line 45: Should we state here that the presence of this partition is conditional? Especially since the option is default `OFF`.

trusted-firmware-m/platform/ext/target/nordic_nrf/common/nrf5340/partition/flash_layout.h
```diff
@@ -30,7 +30,8 @@
...
```
hakonfam, line 34: Could this storage be secure? The commit message states that "If newly added CMake variable NS_STORAGE=ON then configure  the storage area to be non-secure." To me that implies that this area could also be secure storage.

trusted-firmware-m/platform/ext/target/nordic_nrf/common/nrf5340/partition/flash_layout.h
```diff
@@ -148,6 +150,11 @@
                                          FLASH_ITS_AREA_SIZE)
 #define FLASH_NV_COUNTERS_AREA_SIZE     (FLASH_AREA_IMAGE_SECTOR_SIZE)
 
+/* Non-secure storage region */
+#define NRF_FLASH_NS_STORAGE_AREA_OFFSET    (FLASH_TOTAL_SIZE - \
```
hakonfam, line 154: What is the motivation for placing this at the end of flash instead of directly behind the `NV_COUNTER` area? I assume there is some motivation since the placement of this area differs from all the others.

hakonfam (changes requested):
Please post PR upstream and change to a `fromlist`.

The commit message states:

```
tfm: make storage partition available for non-secure world 

Mark the flash region of the storage partition as NS in
SPU. Also, make sure that the partition aligned to the size
of a single flash region in SPU.

```

So to me it sounds like there is an existing partition which, prior to this commit, is only available for secure. Then this commit adds an option which allows you to say that this existing partition should actually be available for non-secure as well.

However, it seems like we are adding a whole new partition, irrespective of any existing secure-only partition.

Have I misunderstood something? If not the commit message should be updated. 

Also, is there supposed to be explicit checks of the size of the partition to ensure that it aligns with the SPU flash config size? Or does the note in the commit message only mean that the size chosen as default is indeed SPU flash config size aligned.

trusted-firmware-m/platform/ext/target/nordic_nrf/common/nrf5340/config.cmake
Damian-Nordic: The upstream TF-M doesn't know the concept of the partition manager etc. It just defines fixed memory regions which are either marked as secure or non-secure. This option basically adds another such a region. The region is hard-code in the upstream to match the storage partition from the device tree, and it's replaced with the actual storage partition (defined with the partition manager) in the NCS.

trusted-firmware-m/platform/ext/target/nordic_nrf/common/nrf5340/partition/flash_layout.h
Damian-Nordic: Yes, good idea.

trusted-firmware-m/platform/ext/target/nordic_nrf/common/nrf5340/partition/flash_layout.h
Damian-Nordic: Ok, I can restate it to "then add a non-secure region for the storage"

trusted-firmware-m/platform/ext/target/nordic_nrf/common/nrf5340/partition/flash_layout.h
Damian-Nordic: Because the device tree for our chips places the storage partition at the end of the flash.

Damian-Nordic:
> Please post PR upstream and change to a `fromlist`.
> 
> The commit message states:
> 
> ```
> tfm: make storage partition available for non-secure world 
> 
> Mark the flash region of the storage partition as NS in
> SPU. Also, make sure that the partition aligned to the size
> of a single flash region in SPU.
> ```
> 
> So to me it sounds like there is an existing partition which, prior to this commit, is only available for secure. Then this commit adds an option which allows you to say that this existing partition should actually be available for non-secure as well.
> 
> However, it seems like we are adding a whole new partition, irrespective of any existing secure-only partition.
> 
> Have I misunderstood something? If not the commit message should be updated.
> 
> Also, is there supposed to be explicit checks of the size of the partition to ensure that it aligns with the SPU flash config size? Or does the note in the commit message only mean that the size chosen as default is indeed SPU flash config size aligned.

This commit message comes from a different PR: https://github.com/nrfconnect/sdk-nrf/pull/4614 which in fact marks the existing storage partition (defined in the partition manager) as non-secure and makes sure that the partition is properly aligned.

hakonfam:
> This commit message comes from a different PR: [nrfconnect/sdk-nrf#4614](https://github.com/nrfconnect/sdk-nrf/pull/4614) which in fact marks the existing storage partition (defined in the partition manager) as non-secure and makes sure that the partition is properly aligned.

Ok, so this will be fixed when this PR is changed to a `fromlist` and the commit message is updated then. Good 👍🏻

Damian-Nordic:
@hakonfam Added an upstream PR: https://github.com/zephyrproject-rtos/trusted-firmware-m/pull/56/files

hakonfam (approved):

Damian-Nordic:
@ioannisg Do you have any comments?

trusted-firmware-m/platform/CMakeLists.txt
```diff
@@ -149,6 +149,7 @@ target_compile_definitions(platform_region_defs
     INTERFACE
         $<$<BOOL:${BL2}>:BL2>
         $<$<BOOL:${SECURE_UART1}>:SECURE_UART1>
+        $<$<BOOL:${NRF_NS_STORAGE}>:NRF_NS_STORAGE>
```
ioannisg, line 152: Ideally this should be moved to nrf-specific CMakeLists.txt file, otherwise, this needs to be upstream-ed.

ioannisg: This is because this is not a generic platform-level option, but rather a Nordic-specific.

trusted-firmware-m/platform/ext/target/nordic_nrf/common/nrf5340/config.cmake
ioannisg: This looks as good as it can be.

What we really need is for upstream tf-m to have a way to define multiple non-secure flash partitions. Currently, TF-M assumes there is a single continuous flash area that is used by the non-secure image.

The solution proposed here is definitely not scalable. In future scenarios there might be even more areas that need to be set as non-secure. What we could possibly do is to use (system) device tree for this, and represent there all flash map configuration with (possibly) multiple secure or non-secure areas.
CC @mbolivar-nordic

trusted-firmware-m/platform/ext/target/nordic_nrf/common/nrf5340/partition/flash_layout.h
```diff
@@ -30,7 +30,8 @@
  * 0x000f_0000 Protected Storage Area (16 KB)
  * 0x000f_4000 Internal Trusted Storage Area (8 KB)
  * 0x000f_6000 NV counters area (4 KB)
- * 0x000f_7000 Unused
+ * 0x000f_7000 Unused (4 KB)
+ * 0x000f_8000 Non-secure storage used when built with NRF_NS_STORAGE=ON (32 KB)
```
ioannisg, line 34: ...otherwise unused?

trusted-firmware-m/platform/ext/target/nordic_nrf/common/nrf5340/partition/flash_layout.h
```diff
@@ -40,7 +41,8 @@
...
```
ioannisg, line 45: ...otherwise is this also unused?

trusted-firmware-m/platform/ext/target/nordic_nrf/common/nrf9160/partition/flash_layout.h
```diff
@@ -30,7 +30,8 @@
  * 0x000f_0000 Protected Storage Area (20 KB)
  * 0x000f_4000 Internal Trusted Storage Area (16 KB)
  * 0x000f_6000 NV counters area (4 KB)
- * 0x000f_7000 Unused
+ * 0x000f_7000 Unused (4 KB)
+ * 0x000f_8000 Non-secure storage used when built with NRF_NS_STORAGE=ON (32 KB)
```
ioannisg, line 34: same comment/question as above

ioannisg (changes requested):
Left some comments. Overall looks good to me.
But this needs to be upstream-ed. Nothing here looks NCS specific.

Damian-Nordic:
@ioannisg @SebastianBoe The upstream PR has been merged, so we should be good to go with this downstream patch as well. Please review :).

SebastianBoe (approved):

PR #19, opened 2021-06-01
tfm: main.c: Provision random HUKs if none are present

oyvindronningstad:
For both IPC and library mode.
Using the hw_unique_key library.

Signed-off-by: Øyvind Rønningstad <oyvind.ronningstad@nordicsemi.no>

### DIFF
--- a/trusted-firmware-m/secure_fw/spm/cmsis_func/main.c
+++ b/trusted-firmware-m/secure_fw/spm/cmsis_func/main.c
@@ -18,6 +18,7 @@
 #include "tfm_spm_hal.h"
 #include "tfm_spm_log.h"
 #include "tfm_version.h"
+#include "hw_unique_key.h"
 
 /*
  * Avoids the semihosting issue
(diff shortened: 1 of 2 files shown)


trusted-firmware-m/secure_fw/spm/cmsis_func/main.c
```diff
@@ -150,6 +151,10 @@ int main(void)
 
     fih_delay_init();
 
+    if (!hw_unique_key_are_any_written()) {
```
ioannisg, line 154: Does this need to be invoked before tfm_core_init(), which is supposed to do a lot of core initialization stuff?

ioannisg (commented):
@oyvindronningstad would it be possible to make this upstreamable?
In the sense that, could it be possible for the hw_unique_key to be  written in our platform code?
Or add a hook here and make this generic enough for any platform?

If so, I'd favor an upstream patch instead.

Should this be a downstream only change, i think it deserves some inline commenting for ease of understanding. Oh, and a git sauce tag, e.g. "noup" :)

ioannisg:
Update: 
Discussed that offline with @oyvindronningstad 
We should try to see if it is possible to move the HUK initialization a little bit further down in tfm_core_init() and in particular in platform init code, so it can be part of the out-of-tree platform support, and reside in sdk-nrf, instead.

oyvindronningstad:
Solved in sdk-nrf

PR #21, opened 2021-06-06
[nrf toup] add option to disable nv_counter

hakonfam:

### DIFF
--- a/trusted-firmware-m/config/config_default.cmake
+++ b/trusted-firmware-m/config/config_default.cmake
@@ -132,6 +132,7 @@ set(ATTEST_INCLUDE_OPTIONAL_CLAIMS      ON          CACHE BOOL      "Include opt
 set(ATTEST_INCLUDE_COSE_KEY_ID          OFF         CACHE BOOL      "Include COSE key-id in initial attestation token")
 
 set(TFM_PARTITION_PLATFORM              ON          CACHE BOOL      "Enable Platform partition")
+set(PLATFORM_NV_COUNTER_MODULE_DISABLED FALSE       CACHE BOOL      "Disable Non-volatile counter module")
 
 set(TFM_PARTITION_AUDIT_LOG             ON          CACHE BOOL      "Enable Audit Log partition")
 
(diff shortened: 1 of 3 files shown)


hakonfam:
See NCSDK-9901

oyvindronningstad:
I would prefer to see what upstream says about this before merging it, since I'm not really familiar with NV counters. We're not using "proper" NV counters, but I don't know whether it's still used in regular flash.

hakonfam:
I see your point. I guess the most important item we need to agree on is that building a TF-M image without support for nv_counters makes sense. IMO it does, since we don't need that in SPM today. Also, since I'm able to remove the functionality, and still build a working TF-M secure image I consider it "safe" in the configuration we use.

oyvindronningstad:
I checked now, and it seems it is only provided as an API, not used by other partitions. So I think this is safe enough.

trusted-firmware-m/config/config_default.cmake
```diff
@@ -132,6 +132,7 @@ set(ATTEST_INCLUDE_OPTIONAL_CLAIMS      ON          CACHE BOOL      "Include opt
 set(ATTEST_INCLUDE_COSE_KEY_ID          OFF         CACHE BOOL      "Include COSE key-id in initial attestation token")
 
 set(TFM_PARTITION_PLATFORM              ON          CACHE BOOL      "Enable Platform partition")
+set(PLATFORM_NV_COUNTER_MODULE_DISABLED FALSE       CACHE BOOL      "Disable Non-volatile counter module")
```
oyvindronningstad, line 135: I don't personally like the _DISABLED negative polarity. I realize it's convention though. Take it as you will :D

oyvindronningstad (approved):

trusted-firmware-m/config/config_default.cmake
hakonfam: Hehe, I agree. But sticking with convention wins here IMO

tejlmand (approved):

ioannisg (approved):

PR #23, opened 2021-06-14
[nrf fromlist] partitions: crypto: introduce option to disable RNG ... 

hakonfam:
Allow users to enable RNG without enabling key derivation.

Ref: NCSDK-100031

### DIFF
--- a/trusted-firmware-m/secure_fw/partitions/crypto/tfm_crypto_secure_api.c
+++ b/trusted-firmware-m/secure_fw/partitions/crypto/tfm_crypto_secure_api.c
@@ -1703,7 +1703,7 @@ psa_status_t psa_key_derivation_key_agreement(
 psa_status_t psa_generate_random(uint8_t *output,
                                  size_t output_size)
 {
-#ifdef TFM_CRYPTO_GENERATOR_MODULE_DISABLED
+#ifdef TFM_CRYPTO_RNG_MODULE_DISABLED
     return PSA_ERROR_NOT_SUPPORTED;
 #else
     psa_status_t status;
@@ -1735,7 +1735,7 @@ psa_status_t psa_generate_random(uint8_t *output,
 #endif
 
     return status;
-#endif /* TFM_CRYPTO_GENERATOR_MODULE_DISABLED */
+#endif /* TFM_CRYPTO_RNG_MODULE_DISABLED */
 }
 
 psa_status_t psa_generate_key(const psa_key_attributes_t *attributes,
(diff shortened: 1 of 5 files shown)


hakonfam:
I will add some motivation to the commit message

oyvindronningstad (approved):

ioannisg (approved):
Fine with this. @oyvindronningstad i am just wondering whether the absence of RNG is not an allowed PSA configuration

oyvindronningstad:
@ioannisg It would anyway be removed if crypto is disabled. Maybe @frkv knows?

hakonfam:
RNG will already be removed if you remove the KEY module

frkv:
I don't know if we are certifiable with RNG disabled. It is required in scope of key provisioning and to support the mandatory PSA Attestation service (to generate boot seed)

It is also a heavy limitation on RTOS when you don't have a working entropy/PRNG driver so I am unsure what the intention is here...

ioannisg:
> f we are certifiable with RNG disabled. It is required in scope of key provisioning and to support the mandatory PSA Attestation service (to generate boot seed)

Agreed about the principle, but as far as I understood, this PR is a "no-op". We are already able to disable the RNG API if we disable the KEY_DERIVATION option. All this PR does is to have separate options for RNG and KEY_DERIVATION. Effectively it allows you to have RNG without key derivation. Correct me if I am missing anything @hakonfam @oyvindronningstad

oyvindronningstad:
@ioannisg That is correct AFAIK.

ioannisg:
change this to a From-List now, @hakonfam  :)

frkv (approved):
LGTM

trusted-firmware-m/secure_fw/partitions/crypto/tfm_crypto_secure_api.c
```diff
@@ -1703,7 +1703,7 @@ psa_status_t psa_key_derivation_key_agreement(
 psa_status_t psa_generate_random(uint8_t *output,
                                  size_t output_size)
 {
-#ifdef TFM_CRYPTO_GENERATOR_MODULE_DISABLED
+#ifdef TFM_CRYPTO_RNG_MODULE_DISABLED
```
ioannisg, line 1706: just a note @hakonfam @oyvindronningstad , this is different upstream, now, so we will need to attend this on next upmerge.

ioannisg:
This is actually merged in upstream TFM now

PR #28, opened 2021-08-06
[nrf fromlist] Build: add internal option for configuring type of ...

hakonfam:
Upstream PR:
https://review.trustedfirmware.org/c/TF-M/trusted-firmware-m/+/10960

This is helpful when a generated file is being used and the
location of the generated file is not known when performing
the configure stage.

Change-Id: Ia8f45cb854b15ee643020fe45122fe0d18cceb0c

### DIFF
--- a/trusted-firmware-m/config/config_default.cmake
+++ b/trusted-firmware-m/config/config_default.cmake
@@ -167,7 +167,14 @@ set(MBEDCRYPTO_PATH                     "DOWNLOAD"  CACHE PATH      "Path to Mbe
 set(MBEDCRYPTO_VERSION                  "mbedtls-2.25.0" CACHE STRING "The version of Mbed Crypto to use")
 set(MBEDCRYPTO_GIT_REMOTE               "https://github.com/ARMmbed/mbedtls.git" CACHE STRING "The URL (or path) to retrieve MbedTLS from.")
 set(MBEDCRYPTO_BUILD_TYPE               "${CMAKE_BUILD_TYPE}" CACHE STRING "Build type of Mbed Crypto library")
-set(TFM_MBEDCRYPTO_CONFIG_PATH          "${CMAKE_SOURCE_DIR}/lib/ext/mbedcrypto/mbedcrypto_config/tfm_mbedcrypto_config_default.h" CACHE STRING "Config to use for Mbed Crypto")
+# Below is the cache type for TFM_MBEDCRYPTO_CONFIG_PATH.
+# This is intended to be controlled by an external build system to support relative path.
+# Relative paths can be supported with one of:
+# - set(TFM_MBEDCRYPTO_CONFIG_TYPE "STRING" CACHE INTERNAL "Internal setting")
+# - -DTFM_MBEDCRYPTO_CONFIG_TYPE=STRING
+# before config_default.cmake is loaded.
+set(TFM_MBEDCRYPTO_CONFIG_TYPE          "PATH" CACHE STRING "Type of the *_MBEDCRYPTO_CONFIG_PATH variables, used internally")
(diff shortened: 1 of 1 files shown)


trusted-firmware-m/config/config_default.cmake
```diff
@@ -167,7 +167,8 @@ set(MBEDCRYPTO_PATH                     "DOWNLOAD"  CACHE PATH      "Path to Mbe
 set(MBEDCRYPTO_VERSION                  "mbedtls-2.25.0" CACHE STRING "The version of Mbed Crypto to use")
 set(MBEDCRYPTO_GIT_REMOTE               "https://github.com/ARMmbed/mbedtls.git" CACHE STRING "The URL (or path) to retrieve MbedTLS from.")
 set(MBEDCRYPTO_BUILD_TYPE               "${CMAKE_BUILD_TYPE}" CACHE STRING "Build type of Mbed Crypto library")
-set(TFM_MBEDCRYPTO_CONFIG_PATH          "${CMAKE_SOURCE_DIR}/lib/ext/mbedcrypto/mbedcrypto_config/tfm_mbedcrypto_config_default.h" CACHE STRING "Config to use for Mbed Crypto")
+set(TFM_MBEDCRYPTO_CONFIG_TYPE          "PATH" CACHE INTERNAL "Type of the *_MBEDCRYPTO_CONFIG_PATH variables")
```
tejlmand, line 170: Are you sure `INTERNAL` at this location is correct ?
From the docs:
> INTERNAL
>
>   ..... Use of this type implies FORCE.

so I would think this would result in this line overruling our setting applied with `-DTFM_MBEDCRYPTO_CONFIG_TYPE=STRING`.

When I said `INTERNAL` I meant that other code could set this as INTERNAL to ensure its value, like:
```
set(TFM_MBEDCRYPTO_CONFIG_TYPE "STRING" CACHE INTERNAL "Config type in NCS is STRING")
.....
# Code that will later do:
include(trusted-firmware-m/config/config_default.cmake)
```
(as we use `-D` to set this value, then it doesn't really matter to us.)

But in this code INTERNAL was not intended.

Please correct me if this code actually work as you intended.

```suggestion
# Below is the cache type for TFM_MBEDCRYPTO_CONFIG_PATH.
# This is intended to be controlled by an external build system to support relative path.
# Relative paths can be supported with one of:
# - set(TFM_MBEDCRYPTO_CONFIG_TYPE "STRING" CACHE INTERNAL "Internal setting")
# - -DTFM_MBEDCRYPTO_CONFIG_TYPE=STRING
# before config_default.cmake is loaded.
set(TFM_MBEDCRYPTO_CONFIG_TYPE "PATH" CACHE STRING "Type of the *_MBEDCRYPTO_CONFIG_PATH variables, used internally")
```

tejlmand (changes requested):
did you test this ?

tejlmand:
Great solution to this problem 👍

trusted-firmware-m/config/config_default.cmake
hakonfam: changed to `STRING` (y)

mbolivar-nordic:
This doesn't apply to my mbolivar-nordic/upmerge-2021-07-26 branch

mbolivar-nordic:
This is a small patch and I would normally just resolve conflicts myself, but I'm not sure @tejlmand is convinced about all the details, so I'm going to hold off for now. Once you agree on the details it would be good to get something that applies cleanly to hte upmerge branch.

hakonfam:
https://github.com/mbolivar-nordic/sdk-trusted-firmware-m/pull/1

tejlmand (approved):

mbolivar-nordic:
> [mbolivar-nordic#1](https://github.com/mbolivar-nordic/sdk-trusted-firmware-m/pull/1)

Merged; thanks

PR #33, opened 2021-12-01
Ncs 1.8 rebase

mbolivar-nordic:
Empty diff with v1.3.99-ncs3-rc1, but there is a `toup` patch that needs to be sent upstream (@hakonfam can you please take care of this?)


SebastianBoe:
@mbolivar-nordic : Talked to Håkon. I will send it upstream.

PR #49, opened 2022-05-27
[nrf toup] platform: nordic_nrf: Add option to initialize hardware to…

joerchan:
… reset on boot

Add an option to the nordic platforms to initialize the hardware state
to the its reset values when the TF-M firmware is booting.
The option is useful when the TF-M firmware image is chain-loaded, for
example, a bootloader, and we need to guarantee that the internal
states of the architecture core blocks are restored to the reset values.

Add option for TF-M core panic, to choose between halting or resetting.
Send the TF-M exception handlers to the access violation handler, similar to how other platform specific handlers do.
And then call to core panic.
In this way all TF-M error handling go to the same reset or halt system behavior.


platform/ext/target/nordic_nrf/common/core/hw_init.c
```diff
@@ -0,0 +1,55 @@
...
+void hw_init_reset_on_boot(void)
+{
+	/* Disable interrupts */
+	__disable_irq();
+
+	/* Reset exception and interrupt mask state (PRIMASK handled by
+	 * __enable_irq below)
+	 */
+	__set_FAULTMASK(0);
+	__set_BASEPRI(0);
```
joerchan, line 36: @carlescufi Zephyr has similar initialization handling, however it does not handle the BASEPRI.
This likely does not result in any issues for zephyr at the moment because of zephyr actually writing this register during its irq_lock/unlock functions.

However that could be a potential problem in the future, so fixing this in zephyr would be wise IMO.

oyvindronningstad (approved):
FYI there are files for this in MCUboot: `bootloader/mcuboot/boot/zephyr/arm_cleanup.c` and `bootloader/mcuboot/boot/zephyr/nrf_cleanup.c`. The last one does cleanup of peripherals if that's of interest.

platform/ext/target/nordic_nrf/common/core/hw_init.c
carlescufi: Are you sure that Zephyr doesn't initialize `BASEPRI` at boot? Can you link me to the place where it should be initialized?

joerchan: Turns out it is initialised here:
https://github.com/zephyrproject-rtos/zephyr/blob/main/arch/arm/core/aarch32/cortex_m/reset.S#L114

But not part of the CONFIG_INIT_ARCH_HW_AT_BOOT option.

joerchan:
> FYI there are files for this in MCUboot: `bootloader/mcuboot/boot/zephyr/arm_cleanup.c` and `bootloader/mcuboot/boot/zephyr/nrf_cleanup.c`. The last one does cleanup of peripherals if that's of interest.

Yes, we do enable this by default when building MCUBoot with TF-M. Which is much cleaner as you can know which HW peripherals was used by the bootloader and disable them.

I'm not sure if the peripheral cleanup is needed. This can be handled by the application/zephyr in the driver before starting to use the peripheral.

frkv (approved):
LGTM

platform/ext/target/nordic_nrf/common/core/CMakeLists.txt
```diff
@@ -66,6 +66,7 @@ target_sources(platform_s
         cmsis_drivers/Driver_Flash.c
         ${HAL_NORDIC_PATH}/nrfx/drivers/src/nrfx_nvmc.c
         nrfx_glue.c
+        hw_init.c
```
hakonfam, line 69: What does this do compared to what we do on line 84?

platform/ext/target/nordic_nrf/common/core/hw_init.c
```diff
@@ -0,0 +1,55 @@
...
+static void hw_init_mpu(void)
+{
+	int i;
+
+	int num_regions =
+		((MPU->TYPE & MPU_TYPE_DREGION_Msk) >> MPU_TYPE_DREGION_Pos);
+
+	for (i = 0; i < num_regions; i++) {
+		ARM_MPU_ClrRegion(i);
+	}
```
hakonfam, line 21: nit: `num_regions` can be `const` and `i` can be declared inside the for loop.

platform/ext/target/nordic_nrf/common/core/hw_init.c
```diff
@@ -0,0 +1,55 @@
...
+#include <nrf.h>
+#include "array.h"
+
+/* This routine clears all ARM MPU region configuration. */
+static void hw_init_mpu(void)
+{
+	int i;
+
+	int num_regions =
+		((MPU->TYPE & MPU_TYPE_DREGION_Msk) >> MPU_TYPE_DREGION_Pos);
```
hakonfam, line 17: Should we comment why this gives the number of regions?

platform/ext/target/nordic_nrf/common/core/hw_init.c
```diff
@@ -0,0 +1,55 @@
...
+	 * __enable_irq below)
+	 */
+	__set_FAULTMASK(0);
+	__set_BASEPRI(0);
+
+	/* Clear MPU region configuration */
+	hw_init_mpu();
+
+	/* Disable NVIC interrupts */
+	for (uint8_t i = 0; i < ARRAY_SIZE(NVIC->ICER); i++) {
```
hakonfam, line 42: nit: use `int` instead of `uint8_t` to make it faster

platform/ext/target/nordic_nrf/common/core/hw_init.c
```diff
@@ -0,0 +1,55 @@
...
+		NVIC->ICER[i] = 0xFFFFFFFF;
+	}
+	/* Clear pending NVIC interrupts */
+	for (uint8_t i = 0; i < ARRAY_SIZE(NVIC->ICPR); i++) {
```
hakonfam, line 46: same as above

platform/ext/target/nordic_nrf/common/core/hw_init.h
```diff
@@ -0,0 +1,13 @@
...
+ * Copyright (c) 2022 Nordic Semiconductor ASA
+ *
+ * SPDX-License-Identifier: LicenseRef-Nordic-5-Clause
+ */
+
+#ifndef HW_INIT_H__
+#define HW_INIT_H__
+
+#define _SCS_BASE_ADDR 0xE000E000
+#define _SCS_MPU_CTRL (_SCS_BASE_ADDR + 0xd94)
```
hakonfam, line 11: These are prefixed with `_` and look internal. Do we need them in the headerfile?

Also, why isn't hw_init_reset_on_boot declared here?

config/build_type/debug.cmake
```diff
@@ -8,3 +8,5 @@
 set(MBEDCRYPTO_BUILD_TYPE               relwithdebinfo CACHE STRING "Build type of Mbed Crypto library")
 set(TFM_SPM_LOG_LEVEL                   TFM_SPM_LOG_LEVEL_DEBUG CACHE STRING "Set debug SPM log level as Debug level")
 set(TFM_PARTITION_LOG_LEVEL             TFM_PARTITION_LOG_LEVEL_DEBUG CACHE STRING "Set debug SP log level as Debug level")
+
+set(CONFIG_TFM_HALT_ON_CORE_PANIC       ON         CACHE BOOL      "On fatal errors in the secure firmware, halt instead of rebooting.")
```
hakonfam, line 12: Do we typically prefix these with `CONFIG_` ? I guess its to simplify the wrapper code in zephyr/ncs, but are we doing it elsewhere?

joerchan: Not typically no, but TF-M project has decided to use `CONFIG_` on new configurations added.

platform/ext/target/nordic_nrf/common/core/hw_init.h
joerchan: Yeah, so this header is only used by assembly. Which is why it only defines some constants from CMSIS. And does not declare the function.

PR #51, opened 2022-06-03
TF-M: Reset nRF peripherals during boot

joerchan:
Reset nRF peripherals during boot.
This is required when using an MCUBoot version without CONFIG_MCUBOOT_NRF_CLEANUP_PERIPHERAL enabled.
This is by default enabled starting from NCS 1.4.

### DIFF
--- a/platform/ext/target/nordic_nrf/common/core/CMakeLists.txt
+++ b/platform/ext/target/nordic_nrf/common/core/CMakeLists.txt
@@ -86,6 +86,7 @@ if (NRF_HW_INIT_RESET_ON_BOOT)
     target_compile_definitions(platform_s
         PUBLIC
             NRF_HW_INIT_RESET_ON_BOOT
+            $<$<BOOL:${NRF_HW_INIT_NRF_PERIPHERALS}>:NRF_HW_INIT_NRF_PERIPHERALS>
     )
 endif()
 
--- a/platform/ext/target/nordic_nrf/common/core/hw_init.c
+++ b/platform/ext/target/nordic_nrf/common/core/hw_init.c
@@ -8,6 +8,13 @@
 #include <nrf.h>
 #include "array.h"
 
+#include <string.h>
+
+#include <hal/nrf_rtc.h>
+#include <hal/nrf_uarte.h>
+#include <hal/nrf_clock.h>
+#include <hal/nrf_dppi.h>
+
 /* This routine clears all ARM MPU region configuration. */
 static void hw_init_mpu(void)
 {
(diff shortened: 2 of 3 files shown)


platform/ext/target/nordic_nrf/common/core/CMakeLists.txt
```diff
@@ -86,6 +86,7 @@ if (NRF_HW_INIT_RESET_ON_BOOT)
     target_compile_definitions(platform_s
         PUBLIC
             NRF_HW_INIT_RESET_ON_BOOT
+            $<$<BOOL:${NRF_HW_INIT_NRF_PERIPHERALS}>:NRF_HW_INIT_NRF_PERIPHERALS>
```
balaji-nordic, line 89: not very clear what this does. but the rest looks fine

balaji-nordic (approved):

platform/ext/target/nordic_nrf/common/core/CMakeLists.txt
sigvartmh: It's a CMake generator expression if you are familiar with it?

Ref. https://gitlab.kitware.com/cmake/cmake/-/issues/18549#note_479462  for explanation why this is done.

balaji-nordic: ahah. thanks for that.

frkv (approved):
LGTM

platform/ext/target/nordic_nrf/common/core/hw_init.c
```diff
@@ -19,6 +26,78 @@ static void hw_init_mpu(void)
...
+    nrf_cleanup_rtc(NRF_RTC1);
+#endif
+#if defined(NRF_RTC2)
+    nrf_cleanup_rtc(NRF_RTC2);
+#endif
+
+#if defined(NRF_UARTE0)
+    nrf_cleanup_uarte(NRF_UARTE0);
+#endif
+#if defined(NRF_UARTE1)
```
sigvartmh, line 91: Generally looks good, I just wonder where these symbols come from? `NRF_UARTE1` etc. 
Is there any mapping between these and what is set for MCUBoot or does this just generally reset whatever hardware is available on the chip?

sigvartmh (approved):
I don't have any to block on this. I just wondered if in the future maybe we want to coordinate between the MCUBoot build/bootloader booting TF-M and the reset of different peripherals with regards to boot speed etc.

joerchan:
> I don't have any to block on this. I just wondered if in the future maybe we want to coordinate between the MCUBoot build/bootloader booting TF-M and the reset of different peripherals with regards to boot speed etc.

That would make sense to do in MCUBoot, I think in TF-M it is best to be paranoid since you don't know which version of mcuboot you are being booted from.

sigvartmh (approved):

PR #53, opened 2022-06-15
platform: nordic_nrf: Build in only required code

pdunaj:
Don't include test and UART sources into building unless needed.

Signed-off-by: Pawel Dunaj <pawel.dunaj@nordicsemi.no>


joerchan (changes requested):
Blocking this since this introduces a regression in `psa_test` (PSA architecture tests).

joerchan:
This PR will not be needed.

PR #54, opened 2022-06-29
[nrf fromtree] platform: nordic_nrf: Exclude nonsecure storage partition when undefined

joerchan:
Exclude the nonsecure storage partition in nRF platforms when the
storage partition was not defined.
This allows the region to be excluded at build time when the size of the
storage partition is zero.
This simplifies the configuration of the partition in out-of-tree
configuration systems.


Vge0rge (approved):
The change makes sense to me. I think that a single commit will be better for this. Since one commit does the exclusion and the other one does only some renamings. But since it is a fromtree I don't mind keep it like this.

PR #56, opened 2022-08-09
[nrf fromlist] platform: nordic_nrf: Separate secondary partition from BL…

joerchan:
…2 config

Separate the nordic_nrf platform support for including the secondary
image partition from the BL2 configuration.
This is required in order to support two use-cases.
- Using a different bootloader than BL2.
- Supporting the secondary partition in external flash (the SPU should
not configure this partition as non-secure in that case).

Upstream PR:
https://review.trustedfirmware.org/c/TF-M/trusted-firmware-m/+/16414

Change-Id: Icfaa5df972d7c2776b326f4f30e6ea5f3487d35e


jli157:
@joerchan  our project doesn't use BL2 as the bootloader. Instead, we are using Zephyr based MCUBOOT as the bootloader so BL2 is disabled. 

How are the changes from this PR beneficial to the case like what we are working on? `NRF_NS_SECONDARY` is the secondary image of the bootloader no matter it is BL2 or whatever?

joerchan:
@jli157 The type of bootloader does not matter. Enabling the secondary partition independent of the BL2 define is required in order to support the secondary partition being placed in external flash, with BL2 as well.
No change is made to the support of BL2 itself in TF-M,

This PR also fixes the issue that was raised here: https://github.com/nrfconnect/sdk-nrf/pull/7511#issuecomment-1149474095
Where the BL2 define was added to include the secondary partition, but this changed the OTP NV struct layout, causing a problem during firmware update.

So this PR is beneficial to you because, it enables the secondary partition needed for your bootloader without defining other BL2 features, and it enables support for placing the secondary partition in external flash.

platform/ext/target/nordic_nrf/common/core/gcc/nordic_nrf_s.ld
```diff
@@ -638,7 +638,7 @@ VENEERS()
 
     Load$$LR$$LR_NS_PARTITION$$Base = NS_PARTITION_START;
 
-#ifdef BL2
+#ifdef NRF_NS_SECONDARY
     Load$$LR$$LR_SECONDARY_PARTITION$$Base = SECONDARY_PARTITION_START;
 #endif /* BL2 */
```
SebastianBoe, line 643: Forgot to update the comment on this line.

SebastianBoe (changes requested):
Let's do a fromlist instead.

Also, see my comment about a comment.

SebastianBoe (approved):

PR #57, opened 2022-08-10
[nrf fromtree] tfm: Export IOCTL non-secure API source file to install

joerchan:
Add the non-secure API IOCTL functions for the nordic platform to the
set of source files exported in the install folder.
In the case where this is built by an external build system instead of
the platform_ns library then this source file needs to be included in
the non-secure application and its build system.


Vge0rge (approved):
LGTM

frkv (approved):
LGTM

SebastianBoe (approved):

PR #58, opened 2022-08-12
[nrf noup] Add configurable stack size for crypto partition.

joerchan:
Add configurable stack size for the TF-M crypto partition.
This is a temporary method of adding this configuration since the
upstream TF-M implementation for this feature could not be backported
in a clean way.

Should be replaced by the upstream feature in future TF-M upmerge:
https://review.trustedfirmware.org/c/TF-M/trusted-firmware-m/+/16125

### DIFF
--- a/secure_fw/partitions/crypto/tfm_crypto.yaml
+++ b/secure_fw/partitions/crypto/tfm_crypto.yaml
@@ -12,7 +12,7 @@
   "priority": "NORMAL",
   "model": "SFN",
   "entry_init": "tfm_crypto_init",
-  "stack_size": "0x2000",
+  "stack_size": "CRYPTO_STACK_SIZE",
   "secure_functions": [
     {
       "name": "TFM_CRYPTO_GET_KEY_ATTRIBUTES",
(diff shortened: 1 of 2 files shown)


frkv (approved):
LGTM

Vge0rge (approved):

SebastianBoe:
[nrf temphack] is no longer a permitted sauce tag unless I am mistaken.

SebastianBoe:
I think we want to use:



nrf fromlist] | patches submitted to upstream for review, and revisions to them

SebastianBoe (changes requested):
Changes requested

joerchan:
> I think we want to use:
> 
> nrf fromlist] | patches submitted to upstream for review, and revisions to them

It is not submitted upstream. There is a PR in upstream, that I cannot backport, due to the amount of conflicts. Which is why I am doing a temporary hack instead. It should be reverted on the next upmerge.
I can use noup / fromlist instead, but I feel like that is just making it more confusing, temphack is the correct description for it.
Unless we are also not allowing temporary hacks to be done?

SebastianBoe:
I agree that temphack is the correct description, but unfortunately it is no longer permitted.

The second-best description is noup with a commit message with the explanation that you just gave.

joerchan:
Updated commit message to use nrf noup tag.

SebastianBoe (approved):

PR #60, opened 2022-08-26
[nrf toup] platform: nordic_nrf: Replace deprecated API with new one

awojasinski-nordicsemi:
This commit replaces API that became deprecated with the release
of nrfx2.9 - see CHANGELOG in NordicSemiconductor:nrfx repository
https://github.com/NordicSemiconductor/nrfx/blob/v2.9.0/CHANGELOG.md

Signed-off-by: Adam Wojasinski <adam.wojasinski@nordicsemi.no>

### DIFF
--- a/platform/ext/target/lairdconnectivity/common/bl5340/target_cfg.c
+++ b/platform/ext/target/lairdconnectivity/common/bl5340/target_cfg.c
@@ -750,8 +750,8 @@ enum tfm_plat_err_t spu_periph_init_cfg(void)
      * This configuration can be done only from secure code, as otherwise those
      * register fields are not accessible.  That's why it is placed here.
      */
-    nrf_gpio_pin_mcu_select(PIN_XL1, NRF_GPIO_PIN_MCUSEL_PERIPHERAL);
-    nrf_gpio_pin_mcu_select(PIN_XL2, NRF_GPIO_PIN_MCUSEL_PERIPHERAL);
+    nrf_gpio_pin_control_select(PIN_XL1, NRF_GPIO_PIN_SEL_PERIPHERAL);
+    nrf_gpio_pin_control_select(PIN_XL2, NRF_GPIO_PIN_SEL_PERIPHERAL);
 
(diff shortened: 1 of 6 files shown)


joerchan:
Can you put NCSDK-11947 in the commit message? Tracking issue to upstream this change.

awojasinski-nordicsemi:
@joerchan I've also opened a PR in https://github.com/zephyrproject-rtos/trusted-firmware-m/pull/76 I'm not sure if it is needed there as well

joerchan:
> @joerchan I've also opened a PR in [zephyrproject-rtos/trusted-firmware-m#76](https://github.com/zephyrproject-rtos/trusted-firmware-m/pull/76) I'm not sure if it is needed there as well

@awojasinski-nordicsemi I think it would be a hard requirement if the deprecated functions are tagged with `__deprecated` and produces build warnings / errors.
If not then it is possible to skip it, but it is no harm in adding it to upstream zephyr as well, I take it zephyr is using nrfx 2.9.0 as well then.

By upstreaming it I mean to send it to the upstream TF-M repository, which would mean also upgrading the nrfx/ hal_nordic version in TF-M.

joerchan (approved):

awojasinski-nordicsemi:
The process we are currently establishing looks like this:

1. nrfx y.z - new API, old one stays and it is mentioned in doxygen and change log that is deprecated and should be changed
2. nrfx y.(z+1) - old API starts to rise warning during compilation
3. nrfx y.(z+2) - old API is in ifdef guards, it starts to break compilation but by defining symbol you can still get that function and compile with warning
4. nrfx y.(z+3) - old API is completely removed

>  I take it zephyr is using nrfx 2.9.0 as well then.
yes it is using nrfx2.9

Basically, it is worth introducing changes in all repositories that are imported by zephyr, sdk-nrf and so on, but it is not urgent to update revisions ASAP as the code will start to generate warnings/errors at the end of this year or maybe later

SebastianBoe (approved):

SebastianBoe:
DNM until manifest PR is green

PR #62, opened 2022-09-01
[nrf noup] Added workaround for missing Debug symbols in tfm_s image

mswarowsky:
patches submitted to upstream for review, and revisions to them:
https://review.trustedfirmware.org/c/TF-M/trusted-firmware-m/+/16703

The linker removes the debug section of the object files, that are
invoked indirectly via the generated veneer functions. Why this
happens, is not fully clear. This commit adds a workaround that
tells the linker to include the functions that are called via a
veneer function. This is done by generating the linker option
-Wl,--undefined=function_name for all veneer functions and writing
them into a file that then later gets added as linker option to
the tfm_s target.

As mentioned it is a workaround, as the root cause is not fully
clear and might be a bug in the ld linker.

Signed-off-by: Markus Swarowsky <markus.swarowsky@nordicsemi.no>

### DIFF
--- a/secure_fw/CMakeLists.txt
+++ b/secure_fw/CMakeLists.txt
@@ -47,10 +47,28 @@ target_compile_options(tfm_s
         ${COMPILER_CP_FLAG}
 )
 
+# This is a workaround to add debug symbols for the veneer functions of the
+# secure part of TF-M. The problem is that the linker strips away the debug
+# section of the functions called by a veneer function. The root cause is not
+# fully clear, but might be a bug in the linker as it removes the debug
+# sections with the reason of being unused but keeps the actual .text section
+# of the functions. To work around this we instruct the linker to expilicitly
+# keep the, "unused" functions.
+if(EXISTS ${CMAKE_BINARY_DIR}/generated/tools/tfm_veneers_functions.txt)
+    file(STRINGS ${CMAKE_BINARY_DIR}/generated/tools/tfm_veneers_functions.txt veneer_functions)
+else()
+    set(veneer_functions "")
+endif()
+
+foreach(function ${veneer_functions})
+    list(APPEND veneer_link_options -Wl,--undefined=${function})
+endforeach()
+
 target_link_options(tfm_s
     PRIVATE
         --entry=Reset_Handler
-        $<$<C_COMPILER_ID:GNU>:-Wl,-Map=${CMAKE_BINARY_DIR}/bin/tfm_s.map>
+        $<$<C_COMPILER_ID:GNU>:-Wl,-Map=${CMAKE_BINARY_DIR}/bin/tfm_s.map> 
+        $<$<C_COMPILER_ID:GNU>:${veneer_link_options}>
         $<$<C_COMPILER_ID:ARMClang>:--map>
         $<$<C_COMPILER_ID:IAR>:--map\;${CMAKE_BINARY_DIR}/bin/tfm_s.map>
     PUBLIC
--- a/tools/tfm_generated_file_list.yaml
+++ b/tools/tfm_generated_file_list.yaml
@@ -34,6 +34,12 @@
         "output": "interface/include/tfm_veneers.h"
     },
     {
+        "name": "Secure Veneers functions file",
+        "short_name": "tfm_veneers_functions",
+        "template": "tools/templates/veneer_function_names.template",
+        "output": "tools/tfm_veneers_functions.txt"
+    },
+    {
         "name": "Secure IRQ handlers",
         "short_name": "tfm_secure_irq_handlers",
         "template": "secure_fw/spm/cmsis_func/tfm_secure_irq_handlers.inc.template",
--- a/tools/templates/veneer_function_names.template
+++ b/tools/templates/veneer_function_names.template
@@ -1,0 +1,5 @@
+{% for partition in partitions %}
+{% for function in partition.manifest.secure_functions%}
+{{function.signal.lower()}}
+{% endfor %}
+{% endfor %}


sigvartmh:
Good find 👯‍♂️

SebastianBoe:
Let's grep for NCSDK-12483 and update anything that references this bug as not fixed.

SebastianBoe:
We need a manifest PR (contact me about how).

tools/templates/veneer_debug_symbols.template
```diff
@@ -0,0 +1,5 @@
+{% for partition in partitions %}
+{% for function in partition.manifest.secure_functions%}
+-Wl,--undefined={{function.signal.lower()}}
```
SebastianBoe, line 3: After reading the docs for --undefined I am surprised that this did not bloat the binary.

Perhaps the binary that was tested already had all veneers?

Can you test/confirm that we can have a veneer in this list without it ending up in the tfm_s.efl?

SebastianBoe (changes requested):
Testing/Clarification requested. Need to do the review in multiple phases unfortunately.

tools/templates/veneer_debug_symbols.template
SebastianBoe: This doesn't add the symbols to the build because whatever is preventing these from being optimized out is already applied.

So there is no risk of bloat we believe.

joerchan:
I think it would be useful to send this patch to upstream TF-M. When given this patch they may either accept it as is, and we can take it as a fromlist/fromtree patch, or they may look deeper into the problem and find a better solution.

We can also do a toup and create a task for upstreaming, but I think we should only do noup if they completely reject the workaround fixing the issue in upstream.

secure_fw/CMakeLists.txt
```diff
@@ -47,10 +47,18 @@ target_compile_options(tfm_s
         ${COMPILER_CP_FLAG}
 )
 
+if(${CMAKE_BUILD_TYPE} STREQUAL "Debug")
+    # To get debug symbols of function called through veneer functions
+    file(STRINGS ${CMAKE_BINARY_DIR}/generated/tools/tfm_veneers_symbols.txt VENEER_SYMBOLS)
```
SebastianBoe, line 52:
```suggestion
    file(STRINGS ${CMAKE_BINARY_DIR}/generated/tools/tfm_veneers_symbols.txt veneer_symbols)
```

lower-case is (generally, not sure about TF-M) used for locally scoped variables like VENEER_SYMBOOLS.

secure_fw/CMakeLists.txt
```diff
@@ -47,10 +47,18 @@ target_compile_options(tfm_s
         ${COMPILER_CP_FLAG}
 )
 
+if(${CMAKE_BUILD_TYPE} STREQUAL "Debug")
```
SebastianBoe, line 50: We want to apply this workaround for all build types.

I want to be able to debug non-Debug builds as well.

e.g. Debug does all kinds of things and may increase the binary size such that I can't use Debug to debug my app. In which case I wish to debug the non-Debug build. And I want debug symbols for this.

tools/tfm_debug_symbol_list.yaml
```diff
@@ -0,0 +1,18 @@
+# This is is just a list for the secure veneer functions, the generated file
```
SebastianBoe, line 1:
```suggestion
# This is just a list for the secure veneer functions, the generated file 
```

tools/tfm_debug_symbol_list.yaml
```diff
@@ -0,0 +1,18 @@
...
+{
+  "name": "TF-M generated veneer symbols",
+  "type": "generated_file_list",
+  "version_major": 0,
+  "version_minor": 1,
+  "file_list": [
+    {
+        "name": "Secure Veneers symbols file",
+        "short_name": "tfm_veneers_c",
+        "template": "tools/templates/veneer_debug_symbols.template",
```
SebastianBoe, line 14:
```suggestion
        "template": "tools/templates/veneer_function_names.template",
```

We can be more specific than "symbols" and (IMHO) since we remove --undefined, we don't need to mention debug specifically as that is only related to how this file is used, and not related to what this file is.

tools/tfm_debug_symbol_list.yaml
```diff
@@ -0,0 +1,18 @@
...
+        "output": "tools/tfm_veneers_symbols.txt"
```
SebastianBoe, line 15:
```suggestion
        "output": "tools/veneer_function_names.txt"
```

tools/templates/veneer_debug_symbols.template
mswarowsky: I tested it with `CONFIG_TFM_CMAKE_BUILD_TYPE_DEBUG=y` and without it, hex size is not changing. 

The rest was discussed offline

tools/CMakeLists.txt
```diff
@@ -160,3 +160,21 @@ if(RET EQUAL 0)
...
+    -f ${CMAKE_CURRENT_SOURCE_DIR}/tfm_debug_symbol_list.yaml
+    -l ${TFM_ISOLATION_LEVEL}
+    -b ${CONFIG_TFM_SPM_BACKEND}
+    -o ${CMAKE_BINARY_DIR}/generated
+    ${PARSE_MANIFEST_QUIET_FLAG}
+    RESULT_VARIABLE RET
+)
+
+if(NOT RET EQUAL 0)
+    message(FATAL_ERROR "Sysmbol File generation failed, debug info maybe missing")
```
SebastianBoe, line 179:
```suggestion
    message(FATAL_ERROR "Symbol File generation failed, debug info may be missing")
```

secure_fw/CMakeLists.txt
```diff
@@ -47,10 +47,18 @@ target_compile_options(tfm_s
...
+    # To get debug symbols of function called through veneer functions
```
SebastianBoe, line 51: Let us give a detailed description of what the problem is and how we are solving it here.

It should include at least everything relevant from the commit message.

It should also include a description of why this does not cause bloat.

And be honest about that we are uncertain of the root cause of the issue and that we suspect a bug in LD.

tools/tfm_debug_symbol_list.yaml
SebastianBoe: Let's replace everywhere in this PR "symbol" with function_name to be more specific.

tools/tfm_debug_symbol_list.yaml
```diff
@@ -0,0 +1,18 @@
+# This is is just a list for the secure veneer functions, the generated file 
+# can be used to add linker instruction to not remove them, so debug info can
+# be provided
```
SebastianBoe, line 3: "See veneer_function_names in secure_fw/CMakeLists.txt for why this is needed."

SebastianBoe (changes requested):

mswarowsky:
> I think it would be useful to send this patch to upstream TF-M. When given this patch they may either accept it as is, and we can take it as a fromlist/fromtree patch, or they may look deeper into the problem and find a better solution.
> 
> We can also do a toup and create a task for upstreaming, but I think we should only do noup if they completely reject the workaround fixing the issue in upstream.

Oversaw the general comment, sorry. 
I was just not expecting anything like this to be going upstream, but if you think it will help, it's worth a try, but honestly not sure about the process

secure_fw/CMakeLists.txt
```diff
@@ -47,10 +47,13 @@ target_compile_options(tfm_s
         ${COMPILER_CP_FLAG}
 )
 
+# To get debug symbols of function called through veneer functions
+file(STRINGS ${CMAKE_BINARY_DIR}/generated/tools/tfm_veneers_symbols.txt VENEER_SYMBOLS)
+
```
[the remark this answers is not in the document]
SebastianBoe: Sorry, bad CMake on my end.

we need list(APPEND veneer_link_options -Wl,--undefined=${function})

secure_fw/CMakeLists.txt
```diff
@@ -47,14 +47,26 @@ target_compile_options(tfm_s
         ${COMPILER_CP_FLAG}
 )
 
+# This is a workaround to add debug symbols for the veneer functions of the
+# secure part of TF-M. The problem is that the linker strips away the debug
+# section of the functions called by a veneer function. The root cause is not
+# fully clear, but might be a bug in the linker as it removes the debug sections
+# with the reason of beeing unused but keeps the actual .text section of the functions
+# This tells the linker to expilicitly keep the, "unused" functions.
```
SebastianBoe, line 55:
```suggestion
# To work around this we instruct the linker to expilicitly keep the, "unused" functions.
```

The word "This" sounds like we are refering to something we just described.

secure_fw/CMakeLists.txt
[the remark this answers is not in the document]
mswarowsky: Oh yeah makes more sense

secure_fw/CMakeLists.txt
```diff
@@ -47,14 +47,27 @@ target_compile_options(tfm_s
         ${COMPILER_CP_FLAG}
 )
 
+# This is a workaround to add debug symbols for the veneer functions of the
+# secure part of TF-M. The problem is that the linker strips away the debug
+# section of the functions called by a veneer function. The root cause is not
+# fully clear, but might be a bug in the linker as it removes the debug 
+# sections with the reason of beeing unused but keeps the actual .text section
```
Vge0rge, line 54:
```suggestion
# sections with the reason of being unused but keeps the actual .text section 
``` 
Lets leave the bees alone

tools/CMakeLists.txt
```diff
@@ -160,3 +160,21 @@ if(RET EQUAL 0)
...
+    -f ${CMAKE_CURRENT_SOURCE_DIR}/tfm_debug_symbol_list.yaml
+    -l ${TFM_ISOLATION_LEVEL}
+    -b ${CONFIG_TFM_SPM_BACKEND}
+    -o ${CMAKE_BINARY_DIR}/generated
+    ${PARSE_MANIFEST_QUIET_FLAG}
+    RESULT_VARIABLE RET
+)
+
+if(NOT RET EQUAL 0)
+    message(FATAL_ERROR "Symbol File generation failed, debug info may be missing")
```
Vge0rge, line 179: I think that the error message does not match with the error level. The error message as I read it implies that the building will continue without symbols. But since you classified it as FATAL_ERROR this will cause the building to stop. I would say better classify it as a warning and keep the error message.

mswarowsky: If this fails, then the file is missing, to read out the functions in secure_fw/CMakeLists.txt, causing then the build to fail. 
but probably there is a way to prevent this from failing.

SebastianBoe: I agree that we can't simply classify as warning as that would just trigger a more obscure error later in the build.

FATAL_ERROR is OK. Although it would be more future-proof to have this configurable and enabled by default in case this machinery causes an unexpected problem and we need a quick workaround to disable it.

A configuration option can be introduced by adding a line to
trusted-firmware-m/config/config_default.cmake

secure_fw/CMakeLists.txt
mswarowsky: But I want more honey...

tools/CMakeLists.txt
mswarowsky: I changed it to a warning, and made the  part for adding the linker option fail-safe if the file doesn't exist

SebastianBoe:
Let's change the sauce tag to:

[nrf fromlist]

"patches submitted to upstream for review, and revisions to them"

And add a link to the upstream review.

mswarowsky:
> Let's change the sauce tag to:
> 
> [nrf fromlist]
> 
> "patches submitted to upstream for review, and revisions to them"
> 
> And add a link to the upstream review.

I changed the PR title and description, or do you also want the commit-msg changed?

SebastianBoe (changes requested):
sauce tag is for the commit message. The PR title is of lesser importance.

It is used when we regularly, using tooling, re-write and clean the history.

tools/tfm_generated_file_list.yaml
```diff
@@ -34,8 +34,12 @@
         "output": "interface/include/tfm_veneers.h"
     },
     {
-        "name": "Secure IRQ handlers",
-        "short_name": "tfm_secure_irq_handlers",
```
joerchan, line 38: Is removing these two a mistake?

secure_fw/CMakeLists.txt
```diff
@@ -47,14 +47,32 @@ target_compile_options(tfm_s
...
+    list(APPEND veneer_link_options -Wl,--undefined=${function})
+endforeach()
+
 target_link_options(tfm_s
     PRIVATE
-        --entry=Reset_Handler
-        $<$<C_COMPILER_ID:GNU>:-Wl,-Map=${CMAKE_BINARY_DIR}/bin/tfm_s.map>
-        $<$<C_COMPILER_ID:ARMClang>:--map>
-        $<$<C_COMPILER_ID:IAR>:--map\;${CMAKE_BINARY_DIR}/bin/tfm_s.map>
+    --entry=Reset_Handler
```
joerchan, line 69: Is it intentional to change the indentation here?
It looks different from the rest of the source file.

mswarowsky: No just and stupid autoformat from my vs code :(

SebastianBoe (approved):
Usually noup commits are not OK.

But in this case we expect the issue to go away when we deprecate library mode right?

So we only need to keep it for a short time, which is acceptable.

secure_fw/CMakeLists.txt
```diff
@@ -47,10 +47,28 @@ target_compile_options(tfm_s
...
+
+foreach(function ${veneer_functions})
+    list(APPEND veneer_link_options -Wl,--undefined=${function})
+endforeach()
+
 target_link_options(tfm_s
     PRIVATE
         --entry=Reset_Handler
-        $<$<C_COMPILER_ID:GNU>:-Wl,-Map=${CMAKE_BINARY_DIR}/bin/tfm_s.map>
+        $<$<C_COMPILER_ID:GNU>:-Wl,-Map=${CMAKE_BINARY_DIR}/bin/tfm_s.map>
```
joerchan, line 70: FYI: There is a trailing whitespace her, might be caught be the checkpatch / nit scripts?

joerchan (approved):

PR #69, opened 2022-10-10
Add support for secure log output on a shared UART instance

joerchan:
Add an option to send the log output from the secure firmware on a
UART instance that would be shared with the non-secure application.

This option is added where the number of UART instances is limited
and the application only cares about the receiving the TF-M log
on fatal errors.

To allow this option to be enabled the log is disabled in the boot
process before the non-secure application is started.
It is enabled again when an unrecoverable exception has occurred in
the secure firmware.

The UART0 instance is configured as non-secure after it has been
uninitialized, and configured as secure when it is initialized again
on a fatal error.


SebastianBoe:
RFC is approved.

config/config_default.cmake
```diff
@@ -61,6 +61,8 @@ set(CONFIG_TFM_LAZY_STACKING            OFF         CACHE BOOL      "Enable/disa
 
 set(CONFIG_TFM_DOORBELL_API             ON          CACHE BOOL      "Enable the doorbell APIs")
 
+set(CONFIG_TFM_LOG_SHARE_UART           OFF         CACHE BOOL      "TF-M shares the UART log instance with the application. Only use it during boot and in fault handling")
+
 ############################ Platform ##########################################
 
 set(TFM_MULTI_CORE_TOPOLOGY             OFF         CACHE BOOL      "Whether to build for a dual-cpu architecture")
```
SebastianBoe, line 68: missing commit sauce tags.

IMHO since this is an old bug it should not be treated as a hotfix and we should do a fromlist instead of toup.

But toup is acceptable if you prefer.

platform/ext/target/nordic_nrf/common/core/cmsis_drivers/Driver_USART.c
```diff
@@ -22,6 +22,11 @@
 #include <nrfx_uarte.h>
 #include <string.h>
 
+#if !(DOMAIN_NS == 1U) && defined(SECURE_UART1) && defined(CONFIG_TFM_LOG_SHARE_UART)
```
SebastianBoe, line 25: Let's put a config check in CMake on CONFIG_TFM_LOG_SHARE_UART requiring SECURE_UART1.

Then we won't accidentally have a config where the UART is supposedly shared, but no logging comes out.

This expression is also simplified.

SebastianBoe (changes requested):

platform/ext/target/nordic_nrf/common/nrf5340/target_cfg.c
```diff
@@ -712,11 +717,19 @@ enum tfm_plat_err_t spu_periph_init_cfg(void)
      * - TWISx
      * - UARTEx
      */
+
+    /* UART0 is a secure peripheral, so we need to leave Serial-Box 0 as Secure.
+     * The UART Driver will configure it as non-secure when it uninitializes.
+     */
+#if !(defined(SECURE_UART1) && defined(CONFIG_TFM_SHARE_UART))
```
Vge0rge, line 724: Since we always require that SECURE_UART1 is not defined it is better to have a top if statement with this and a nested statement for the other configurations.

platform/ext/target/nordic_nrf/common/nrf9160/target_cfg.c
```diff
@@ -581,11 +586,19 @@ enum tfm_plat_err_t spu_periph_init_cfg(void)
      * - TWISx
      * - UARTEx
      */
+
+    /* UART0 is a secure peripheral, so we need to leave Serial-Box 0 as Secure.
+     * The UART Driver will configure it as non-secure when it uninitializes.
+     */
+#if !(defined(SECURE_UART1) && defined(CONFIG_TFM_SHARE_UART))
```
Vge0rge, line 593: Same comment as before, better to have a top level if with SECURE_UART to improve readability.

Vge0rge (approved):
LGTM

platform/ext/target/nordic_nrf/common/nrf5340/target_cfg.c
joerchan: You're missing the parenthesis, the not is for the whole expression. So your suggestion is not possible without duplicating the call to "config_non_secure".

config/config_default.cmake
```diff
@@ -61,6 +61,7 @@ set(CONFIG_TFM_LAZY_STACKING            OFF         CACHE BOOL      "Enable/disa
 
 set(CONFIG_TFM_DOORBELL_API             ON          CACHE BOOL      "Enable the doorbell APIs")
 
+set(CONFIG_TFM_LOG_SHARE_UART           OFF         CACHE BOOL      "Allow TF-M and the non-secure application to share the UART instance. TF-M will use ut during its boot, after which the non-secure application will use it until an eventual fatal error is handled and logged by TF-M. Logging from TF-M will therefore otherwise be suppressed")
```
SebastianBoe, line 64:
```suggestion
set(CONFIG_TFM_LOG_SHARE_UART           OFF         CACHE BOOL      "Allow TF-M and the non-secure application to share the UART instance. TF-M will use it while its booting, after which the non-secure application will use it until an eventual fatal error is handled and logged by TF-M. Logging from TF-M will therefore otherwise be suppressed")
```

platform/ext/target/nordic_nrf/common/nrf5340/target_cfg.c
```diff
@@ -711,11 +716,19 @@ enum tfm_plat_err_t spu_periph_init_cfg(void)
      * - TWISx
      * - UARTEx
      */
+
+    /* When UART0 is a secure peripheral we need to leave Serial-Box 0 as Secure.
+     * The UART Driver will configure it as non-secure when it uninitializes.
+     */
+#if !(defined(SECURE_UART1) && NRF_SECURE_UART_INSTANCE == 0)
```
SebastianBoe, line 723: If it's intentional that this says SECURE_UART1 then it should be explicitly documented why as without context it looks like a copy-paste error.

SebastianBoe (approved):

PR #74, opened 2022-10-31
Do not attempt to reset locked regions

vili-nordic:
With a bootloader enabled some regions are locked and cannot be reset by TF-M. This commit checks whether the region is locked before resetting them.

### DIFF
--- a/platform/ext/target/nordic_nrf/common/core/native_drivers/spu.c
+++ b/platform/ext/target/nordic_nrf/common/core/native_drivers/spu.c
@@ -72,24 +72,77 @@ void spu_clear_events(void)
     nrf_spu_event_clear(NRF_SPU, NRF_SPU_EVENT_PERIPHACCERR);
 }
 
-void spu_regions_reset_all_secure(void)
+#if defined(REGION_MCUBOOT_ADDRESS) || defined(REGION_B0_ADDRESS) || defined(REGION_S0_ADDRESS) || defined(REGION_S1_ADDRESS)
+static bool spu_region_is_flash_region_in_address_range(uint8_t region_id, uint32_t start_address, uint32_t end_address)
+{
+    size_t start_id = (start_address - DEVICE_FLASH_BASE_ADDRESS) / FLASH_SECURE_ATTRIBUTION_REGION_SIZE;
+    size_t end_id = (end_address - DEVICE_FLASH_BASE_ADDRESS) / FLASH_SECURE_ATTRIBUTION_REGION_SIZE;
+    return region_id >= start_id && region_id <= end_id;
+}
+#endif
+
+#if defined(REGION_PCD_SRAM_ADDRESS)
+static bool spu_region_is_sram_region_in_address_range(uint8_t region_id, uint32_t start_address, uint32_t end_address)
+{
+    size_t start_id = (start_address - DEVICE_SRAM_BASE_ADDRESS) / SRAM_SECURE_ATTRIBUTION_REGION_SIZE;
+    size_t end_id = (end_address - DEVICE_SRAM_BASE_ADDRESS) / SRAM_SECURE_ATTRIBUTION_REGION_SIZE;
+    return region_id >= start_id && region_id <= end_id;
+}
+#endif
+
+static bool spu_region_is_bootloader_region(NRF_SPU_Type * p_reg, uint8_t region_id)
+{
+    bool is_bootloader = false;
+
+#ifdef REGION_MCUBOOT_ADDRESS
+    is_bootloader = is_bootloader || spu_region_is_flash_region_in_address_range(region_id, REGION_MCUBOOT_ADDRESS, REGION_MCUBOOT_END_ADDRESS);
+#endif
+#ifdef REGION_B0_ADDRESS
+    is_bootloader = is_bootloader || spu_region_is_flash_region_in_address_range(region_id, REGION_B0_ADDRESS, REGION_B0_END_ADDRESS);
+#endif
+#ifdef REGION_S0_ADDRESS
+    is_bootloader = is_bootloader || spu_region_is_flash_region_in_address_range(region_id, REGION_S0_ADDRESS, REGION_S0_END_ADDRESS);
+#endif
+#ifdef REGION_S1_ADDRESS
+    is_bootloader = is_bootloader || spu_region_is_flash_region_in_address_range(region_id, REGION_S1_ADDRESS, REGION_S1_END_ADDRESS);
+#endif
+
+    return is_bootloader;
+}
+
+static bool spu_region_is_pcd_region(NRF_SPU_Type * p_reg, uint8_t region_id)
+{
+    bool is_pcd = false;
+
+#ifdef PM_PCD_SRAM_ADDRESS
+    is_pcd = is_pcd || spu_region_is_sram_region_in_address_range(region_id, PM_PCD_SRAM_ADDRESS, PM_PCD_SRAM_END_ADDRESS);
+#endif
+
+    return is_pcd;
+}
+
+void spu_regions_reset_unlocked_secure(void)
 {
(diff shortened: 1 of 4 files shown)


platform/ext/target/nordic_nrf/common/core/native_drivers/spu.c
```diff
@@ -72,15 +72,23 @@ void spu_clear_events(void)
...
-            0 /* No lock */);
+        if (!spu_region_is_locked(NRF_SPU, i))
+        {
+            nrf_spu_flashregion_set(NRF_SPU, i,
+                1 /* Secure */,
+                NRF_SPU_MEM_PERM_READ
+                | NRF_SPU_MEM_PERM_WRITE
+                | NRF_SPU_MEM_PERM_EXECUTE,
+                0 /* No lock */);
+            }
```
SebastianBoe, line 91: I feel it is unfortunate that the function name states it is resetting all secure SPU regions while in actuality it is not.

I also feel it is dangerous for TF-M to assume that whatever was configured prior is appropriate.

Some kind of config enforcement, either build/config time or runtime is necessary I think.

Presumably TF-M is resetting all the config prior to doing some config. So it must know what it needs to configure and which regions need resetting.

Perhaps TF-M should reset the regions it intends to configure instead of resetting all regions?

SebastianBoe: Also, patches to upstream projects need to be rationalized:

"This is not upstream because X. It can be removed when Y."

vili-nordic: SPU init essentially does:
* Set all regions as secure (this reset)
* Iterate through non-secure regions and set them up.
It does not specifically know which regions to configure as secure. If the region is locked TF-M cannot configure it to be different anyways (could fail with reset or something though).

I agree this is not nice, and it was also my concern when discussing this @Vge0rge earlier today.

SebastianBoe: The TF-M build has accesss to pm_config.h I believe.

Perhaps it could infer from pm_config.h which SPU regions are assigned to the bootloaders and only skip configuring those regions if they are both locked and in a bootloader region?

vili-nordic: Thanks. I'll look into that

platform/ext/target/nordic_nrf/common/core/native_drivers/spu.c
```diff
@@ -72,15 +73,65 @@ void spu_clear_events(void)
...
+    }
+#endif
+
+#ifdef PM_MCUBOOT_ADDRESS
+    {
+        size_t start_id = (PM_MCUBOOT_ADDRESS - DEVICE_FLASH_BASE_ADDRESS) / FLASH_SECURE_ATTRIBUTION_REGION_SIZE;
+        size_t end_id = (PM_MCUBOOT_END_ADDRESS - DEVICE_FLASH_BASE_ADDRESS) / FLASH_SECURE_ATTRIBUTION_REGION_SIZE;
+        is_bootloader = is_bootloader || (region_id >= start_id && region_id <= end_id);
+    }
+#endif
```
SebastianBoe, line 94: NB: Copy-paste error. PM_MCUBOOT_ADDRESS.

platform/ext/target/nordic_nrf/common/core/native_drivers/spu.c
```diff
@@ -72,15 +73,65 @@ void spu_clear_events(void)
...
+
+#ifdef PM_S1_ADDRESS
+    {
+        size_t start_id = (PM_S1_ADDRESS - DEVICE_FLASH_BASE_ADDRESS) / FLASH_SECURE_ATTRIBUTION_REGION_SIZE;
+        size_t end_id = (PM_S1_END_ADDRESS - DEVICE_FLASH_BASE_ADDRESS) / FLASH_SECURE_ATTRIBUTION_REGION_SIZE;
+        is_bootloader = is_bootloader || (region_id >= start_id && region_id <= end_id);
+    }
+#endif
+
+    return is_bootloader;
```
SebastianBoe, line 120: #ifdef PM_MCUBOOT_ADDRESS
is_bootloader |= some_static_function(region_id, PM_MCUBOOT_ADDRESS, PM_MCUBOOT_END_ADDRESS);
#endif

Let's reduce the amount of copy-paste to avoid copy-paste errors.

platform/ext/target/nordic_nrf/common/core/native_drivers/spu.c
vili-nordic: will fix

platform/ext/target/nordic_nrf/common/core/native_drivers/spu.c
```diff
@@ -72,15 +73,47 @@ void spu_clear_events(void)
     nrf_spu_event_clear(NRF_SPU, NRF_SPU_EVENT_PERIPHACCERR);
 }
 
-void spu_regions_reset_all_secure(void)
+#if defined(PM_MCUBOOT_ADDRESS) || defined(PM_B0_ADDRESS) || defined(PM_S0_ADDRESS) || defined(PM_S1_ADDRESS)
```
SebastianBoe, line 76: What are these ifdef's for?

Are we getting an unused function warning?

platform/ext/target/nordic_nrf/common/nrf9160/target_cfg.c
```diff
@@ -533,7 +533,7 @@ enum tfm_plat_err_t spu_init_cfg(void)
      * in case this has been overwritten by earlier images e.g.
      * bootloader.
      */
-    spu_regions_reset_all_secure();
+    spu_regions_reset_unlocked_secure();
```
SebastianBoe, line 536: The comment above here doesn't make sense anymore.

(I realize it was wrong before, the code was trying to do something and failing, now we are trying to do something else, and the comment must reflect what we are now trying to do)

platform/ext/target/nordic_nrf/common/core/native_drivers/spu.h
```diff
@@ -40,14 +40,14 @@ void spu_enable_interrupts(void);
 void spu_clear_events(void);
 
 /**
- * \brief Reset all memory regions to being Secure
+ * \brief Reset TF-M memory regions to being Secure.
  *
- * Reset all (Flash or SRAM) memory regions to being Secure
- * and have default (i.e. Read-Write-Execute allow) access policy
+ * Reset all (Flash or SRAM) memory regions that have not been locked by any bootloader
+ * to be Secure and have default (i.e. Read-Write-Execute allow) access policy.
```
SebastianBoe, line 46:
```suggestion
 * Reset all (Flash or SRAM) memory region permissions that have not already been locked by a bootloader.
```

It sounds like a bootloader could be locking things as non-secure with the previous wording.

platform/ext/target/nordic_nrf/common/core/native_drivers/spu.c
vili-nordic: yes

SebastianBoe (changes requested):
It's my understanding that we will use region_defs to avoid a noup commit.

I'm keeping this -1 to track that this has not been done yet.

SebastianBoe (changes requested):
Commits must be tagged with nrf saucetags.

https://projecttools.nordicsemi.no/confluence/display/NCS/Pull+Requests

Nearly there :)

SebastianBoe (approved):

joerchan (approved):

nordicjm (changes requested):
Missing PR that this was taken from in commit message

nordicjm (approved):

PR #77, opened 2022-11-22
[nrf fromtree]: Crypto: Make sure output length is used only when PSA_SUCCESS is returned

Vge0rge:
When a PSA Crypto API has a parameter that is used to return the size of
an output buffer, the service needs to make sure that it uses the value
returned only in case the API itself returns PSA_SUCCESS, as there is no
guarantee that the value returned for the size of output buffers will be
meaningful in error cases.

Ref: NCSDK-18172

Signed-off-by: Georgios Vasilakis <georgios.vasilakis@nordicsemi.no>

### DIFF
--- a/secure_fw/partitions/crypto/crypto_aead.c
+++ b/secure_fw/partitions/crypto/crypto_aead.c
@@ -133,8 +133,8 @@ psa_status_t tfm_crypto_aead_encrypt_setup(psa_invec in_vec[],
     psa_algorithm_t alg = iov->alg;
     mbedtls_svc_key_id_t encoded_key;
 
-    /* Set the output handle to invalid, before allocation */
-    *handle_out = TFM_CRYPTO_INVALID_HANDLE;
+    /* Init the handle in the operation with the one passed from the iov */
+    *handle_out = iov->op_handle;
 
     /* Allocate the operation context in the secure world */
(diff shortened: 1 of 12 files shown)


SebastianBoe (approved):

joerchan (approved):

secure_fw/partitions/crypto/crypto_aead.c
```diff
@@ -30,7 +30,7 @@ psa_status_t tfm_crypto_aead_encrypt(psa_invec in_vec[],
 #else
     psa_status_t status = PSA_SUCCESS;
 
-    CRYPTO_IN_OUT_LEN_VALIDATE(in_len, 1, 3, out_len, 0, 1);
+    CRYPTO_IN_OUT_LEN_VALIDATE(in_len, 1, 3, out_len, 1, 1);
```
magnev, line 33: This change makes the PSA API Crypto test number 224 fail.
It fails because the "Unsupported algorithm" test vector sets `ciphertext_size` to 0, and makes the test to fail too early.
Ideally, the test vector should be updated to have correct parameters, except for the unsupported algorithm.

Vge0rge: This was accidentally taken in while resolving merge conflicts. I have now removed it.

PR #79, opened 2022-12-01
Free operation context on setup failure

vili-nordic:
If psa_aead_encrypt_setup or psa_aead_decrypt_setup is given a operation struct that is already in use, free it before returning error.

Signed-off-by: Vidar Lillebø <vidar.lillebo@nordicsemi.no>

### DIFF
--- a/secure_fw/partitions/crypto/crypto_aead.c
+++ b/secure_fw/partitions/crypto/crypto_aead.c
@@ -133,6 +133,13 @@ psa_status_t tfm_crypto_aead_encrypt_setup(psa_invec in_vec[],
     psa_algorithm_t alg = iov->alg;
     mbedtls_svc_key_id_t encoded_key;
 
+    if (handle != TFM_CRYPTO_INVALID_HANDLE)
+    {
+        status = PSA_ERROR_BAD_STATE;
+        *handle_out = handle;
+        goto exit;
+    }
+
     /* Set the output handle to invalid, before allocation */
     *handle_out = TFM_CRYPTO_INVALID_HANDLE;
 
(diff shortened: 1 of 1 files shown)


frkv (approved):
LGTM

Vge0rge (approved):

secure_fw/partitions/crypto/crypto_aead.c
```diff
@@ -133,6 +133,13 @@ psa_status_t tfm_crypto_aead_encrypt_setup(psa_invec in_vec[],
     psa_algorithm_t alg = iov->alg;
     mbedtls_svc_key_id_t encoded_key;
 
+    if (handle != TFM_CRYPTO_INVALID_HANDLE)
+    {
+        status = PSA_ERROR_BAD_STATE;
+        *handle_out = handle;
+        goto exit;
+    }
```
joerchan, line 141: Releasing the allocated handle in this case is not the right fix I think.
Once the previous setup function has returned success it should either go through a successful finish, or an abort function. 
This is the documented behavior: https://armmbed.github.io/mbed-crypto/html/api/ops/aead.html#c.psa_aead_encrypt_setup
> After a successful call to psa_aead_encrypt_setup(), the application must eventually terminate the operation. The following events terminate an operation:
> 
> A successful call to psa_aead_finish().
> A call to psa_aead_abort().

The abort functionality does more validation of the input handle, so this seems like a much safer approach. Blindly releasing the given handle like this seem unsafe.

```suggestion
    if (handle != TFM_CRYPTO_INVALID_HANDLE)
    {
        *handle_out = handle;
        return PSA_ERROR_BAD_STATE;
    }
```

secure_fw/partitions/crypto/crypto_aead.c
```diff
@@ -189,6 +196,13 @@ psa_status_t tfm_crypto_aead_decrypt_setup(psa_invec in_vec[],
...
+
```
joerchan, line 205:
```suggestion
    if (handle != TFM_CRYPTO_INVALID_HANDLE)
    {
        *handle_out = handle;
        return PSA_ERROR_BAD_STATE;
    }
```

joerchan (changes requested):

joerchan:
Can you add a reference to the Jira issue in the commit message? NCSDK-18447?

secure_fw/partitions/crypto/crypto_aead.c
magnev: I agree with @joerchan here.
I tested this proposed solution locally successfully.

PR #82, opened 2023-02-13
Update to TF-M 1.7.0

joerchan:


config/config_base.cmake
```diff
@@ -61,6 +61,7 @@ set(CONFIG_TFM_STACK_WATERMARKS         OFF         CACHE BOOL      "Whether to
 
 set(PROJECT_CONFIG_HEADER_FILE          "${CMAKE_SOURCE_DIR}/config/config_base.h" CACHE FILEPATH "User defined header file for TF-M config")
 
+set(CONFIG_TFM_LOG_SHARE_UART           OFF         CACHE BOOL      "Allow TF-M and the non-secure application to share the UART instance. TF-M will use ut during its boot, after which the non-secure application will use it until an eventual fatal error is handled and logged by TF-M. Logging from TF-M will therefore otherwise be suppressed")
```
mswarowsky: typo: `use **it** during

mswarowsky (approved):

config/config_base.cmake
joerchan: This is actually fixed in the next commit, looks like a wrong fixup. Will change the commits to avoid this.

PR #83, opened 2023-02-20
NCS v2.3 rebase

mbolivar-nordic:
Regular history rebase


tmon-nordic (commented):
This actually removes the revert and sequashes fixup commit on top of the base commit, so there is some value. However, I am not sure if it is worth it considering that the base commit is the same as prior to history rewrite, i.e. there was no mergeup since last history rewrite.

mbolivar-nordic:
As mentioned in other responses, this is our policy.

tmon-nordic (approved):

plskeggs (approved):

PR #85, opened 2023-03-13
[nrf toup]: ITS: Add support for encrypted ITS

mswarowsky:
This adds the support for EITS again, which had to be reverted for the update to TF-M 1.7
As the upstream PR didn't get merged yet.

-Adds encryption and authentication support for ITS files 
-Encryption is optional and is enabled using a CMake variable
-The encryption implementation is platform dependent, the signatures of the functions are provided in this change

Ref: NCSDK-19574

Change-Id: Iea6ed79907f046a79ff7ff44ea2c7285ee14ddab

### DIFF
--- a/config/config_base.cmake
+++ b/config/config_base.cmake
@@ -108,6 +108,9 @@ set(PS_ENCRYPTION                       ON          CACHE BOOL      "Enable encr
 set(PS_CRYPTO_AEAD_ALG                  PSA_ALG_GCM CACHE STRING    "The AEAD algorithm to use for authenticated encryption in Protected Storage")
 
 set(TFM_PARTITION_INTERNAL_TRUSTED_STORAGE OFF      CACHE BOOL      "Enable Internal Trusted Storage partition")
+set(TFM_ITS_ENCRYPTED                   OFF         CACHE BOOL      "Enable authenticated encryption of ITS files using platform specific APIs")
+set(TFM_ITS_AUTH_TAG_LENGTH             "16"        CACHE STRING    "The size of the authentication tag used when authentication/encryption of ITS files is enabled ")
+set(TFM_ITS_ENC_NONCE_LENGTH            "12"        CACHE STRING    "The size of the nonce used when ITS file encryption is enabled")
 
 set(TFM_PARTITION_CRYPTO                OFF         CACHE BOOL      "Enable Crypto partition")
 set(CRYPTO_TFM_BUILTIN_KEYS_DRIVER      ON          CACHE BOOL      "Whether to allow crypto service to store builtin keys. Without this, ALL builtin keys must be stored in a platform-specific location")
--- a/platform/ext/target/nordic_nrf/common/core/tfm_hal_its_encryption.c
+++ b/platform/ext/target/nordic_nrf/common/core/tfm_hal_its_encryption.c
@@ -1,0 +1,307 @@
+/*
+ * Copyright (c) 2023 Nordic Semiconductor ASA.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include "platform/include/tfm_platform_system.h"
+#include "platform/include/tfm_hal_its.h"
+#include "psa/crypto.h"
+#include <string.h>
+#include <stdint>
+
+/* The cc3xx driver is used directly in this file as the PSA APIs depends on
+ * ITS, so using PSA APIs would result in circular dependencies
+ */
+#include "nrf_cc3xx_platform_derived_key.h"
+#include "nrf_cc3xx_platform.h"
+#include "nrf_cc3xx_platform_kmu.h"
+
+#define ITS_ENCRYPTION_SUCCESS 0
+
+#define HUK_KMU_SLOT 2
+#define HUK_KMU_SIZE_BITS 128
+
+/* Global encryption counter which resets per boot. The counter ensures that
+ * the nonce will not be identical for consecutive file writes during the same
+ * boot.
+ */
+static uint32_t g_enc_counter;
+
+/* The global nonce seed which is fetched once in every boot. The seed is used
+ * as part of the nonce and allows the platforms to diversify their nonces
+ * across resets. Note that the way that this seed is generated is platform
+ * specific, so the diversification is optional.
+ */
+static uint8_t g_enc_nonce_seed[TFM_ITS_ENC_NONCE_LENGTH -
+                                sizeof(g_enc_counter)];
+
+/* TFM_ITS_ENC_NONCE_LENGTH is configurable but this implementation expects
+ * the seed to be 8 bytes and the nonce length to be 12.
+ */
+#if TFM_ITS_ENC_NONCE_LENGTH != 12
+#error "This implementation only supports a ITS nonce of size 12"
+#endif
+
+/*
+ * This implementation doesn't use monotonic counters, but therfore a 64 bit
+ * seed combined with a counter, that gets reset on each reboot.
+ * This still has the risk of getting a collision on the seed resulting in
+ * nonce's beeing the same after a reboot.
+*/
+enum tfm_hal_status_t tfm_hal_its_aead_generate_nonce(uint8_t *nonce,
+                                                      size_t nonce_size)
+{
+    int err;
+
+    if(nonce == NULL){
+        return TFM_HAL_ERROR_INVALID_INPUT;
+    }
+
+    if(nonce_size < sizeof(g_enc_nonce_seed) + sizeof(g_enc_counter)){
+        return TFM_HAL_ERROR_INVALID_INPUT;
+    }
+
+    /* To avoid wrap-around of the g_enc_counter and subsequent re-use of the
+     * nonce we check the counter value for its max value
+     */
+    if(g_enc_counter ==  UINT32_MAX) {
+        return TFM_HAL_ERROR_GENERIC;
+    }
+
+    if (g_enc_counter == 0) {
+        err = nrf_cc3xx_platform_get_nonce_seed(g_enc_nonce_seed);
+        if (err != 0) {
+            return TFM_HAL_ERROR_GENERIC;
+        }
+    }
+
+    memcpy(nonce, g_enc_nonce_seed, sizeof(g_enc_nonce_seed));
+    memcpy(nonce + sizeof(g_enc_nonce_seed),
+               &g_enc_counter,
+               sizeof(g_enc_counter));
+
+    g_enc_counter++;
+
+    return TFM_HAL_SUCCESS;
+}
+
+enum tfm_hal_status_t tfm_hal_its_aead_set_deriv_label(
+                                        struct tfm_hal_its_auth_crypt_ctx *ctx,
+                                        uint8_t *deriv_label,
+                                        size_t deriv_label_size)
+{
+    if (ctx == NULL || deriv_label == NULL) {
+        return TFM_HAL_ERROR_INVALID_INPUT;
+    }
+
+    ctx->deriv_label = deriv_label;
+    ctx->deriv_label_size = deriv_label_size;
+
+    return TFM_HAL_SUCCESS;
+}
+
+enum tfm_hal_status_t tfm_hal_its_aead_set_nonce(
+                                        struct tfm_hal_its_auth_crypt_ctx *ctx,
+                                        uint8_t *nonce,
+                                        size_t nonce_size)
+{
+    if (ctx == NULL || nonce == NULL) {
+        return TFM_HAL_ERROR_INVALID_INPUT;
+    }
+
+    ctx->nonce = nonce;
+    ctx->nonce_size = nonce_size;
+
+    return TFM_HAL_SUCCESS;
+}
+
+enum tfm_hal_status_t tfm_hal_its_aead_set_ad(
+                                        struct tfm_hal_its_auth_crypt_ctx *ctx,
+                                        uint8_t *ad,
+                                        size_t ad_size) {
+
+    if (ctx == NULL) {
+        return TFM_HAL_ERROR_INVALID_INPUT;
+    }
+
+    if (ad == NULL && ad_size != 0) {
+        return TFM_HAL_ERROR_INVALID_INPUT;
+    }
+
+    ctx->add = ad;
+    ctx->add_size = ad_size;
+
+    return TFM_HAL_SUCCESS;
+}
+
+
+static bool ctx_is_valid(struct tfm_hal_its_auth_crypt_ctx *ctx)
+{
+    bool ret;
+
+    if (ctx == NULL) {
+        return false;
+    }
+
+    ret = (ctx->deriv_label == NULL && ctx->deriv_label_size != 0) ||
+          (ctx->add == NULL && ctx->add_size != 0) ||
+          (ctx->nonce == NULL && ctx->nonce_size != 0);
+
+    return !ret;
+}
+
+static enum tfm_hal_status_t tfm_hal_its_aead_init(
+                            struct tfm_hal_its_auth_crypt_ctx *ctx,
+                            nrf_cc3xx_platform_derived_key_ctx_t *platform_ctx,
+                            uint8_t *tag,
+                            size_t tag_size)
+{
+
+    int err = NRF_CC3XX_PLATFORM_ERROR_INTERNAL;
+
+
+    err = nrf_cc3xx_platform_derived_key_init(platform_ctx);
+    if (err != NRF_CC3XX_PLATFORM_SUCCESS) {
+        return TFM_HAL_ERROR_GENERIC;
+    }
+
+    err = nrf_cc3xx_platform_derived_key_set_info(platform_ctx,
+                                                  HUK_KMU_SLOT,
+                                                  HUK_KMU_SIZE_BITS,
+                                                  ctx->deriv_label,
+                                                  ctx->deriv_label_size);
+    if (err != NRF_CC3XX_PLATFORM_SUCCESS) {
+        return TFM_HAL_ERROR_INVALID_INPUT;
+    }
+
+    err = nrf_cc3xx_platform_derived_key_set_cipher(platform_ctx,
+                                                    ALG_CHACHAPOLY_256_BIT);
+
+    if (err != NRF_CC3XX_PLATFORM_SUCCESS) {
+        return TFM_HAL_ERROR_INVALID_INPUT;
+    }
+
+    err = nrf_cc3xx_platform_derived_key_set_auth_info(platform_ctx,
+                                                       ctx->nonce,
+                                                       ctx->nonce_size,
+                                                       ctx->add,
+                                                       ctx->add_size,
+                                                       tag,
+                                                       tag_size);
+    if (err != NRF_CC3XX_PLATFORM_SUCCESS) {
+        return TFM_HAL_ERROR_INVALID_INPUT;
+    }
+
+    return TFM_HAL_SUCCESS;
+}
+
+static void tfm_hal_its_aead_cleanup(
+                            nrf_cc3xx_platform_derived_key_ctx_t *platform_ctx)
+{
+    if(platform_ctx != NULL){
+        memset(platform_ctx,
+                   0x0,
+                   sizeof(nrf_cc3xx_platform_derived_key_ctx_t));
+    }
+}
+
+enum tfm_hal_status_t tfm_hal_its_aead_encrypt(
+                                        struct tfm_hal_its_auth_crypt_ctx *ctx,
+                                        uint8_t *plaintext,
+                                        size_t plaintext_size,
+                                        uint8_t *ciphertext,
+                                        size_t ciphertext_size,
+                                        uint8_t *tag,
+                                        size_t tag_size)
+{
+    nrf_cc3xx_platform_derived_key_ctx_t platform_ctx = {0};
+    enum tfm_hal_status_t err = TFM_HAL_ERROR_GENERIC;
+    int plat_err = NRF_CC3XX_PLATFORM_ERROR_INTERNAL;
+
+    if (!ctx_is_valid(ctx) || tag == NULL) {
+        return TFM_HAL_ERROR_INVALID_INPUT;
+    }
+
+    if (plaintext_size > ciphertext_size) {
+        return TFM_HAL_ERROR_INVALID_INPUT;
+    }
+
+    err = tfm_hal_its_aead_init(ctx,
+                                &platform_ctx,
+                                tag,
+                                tag_size);
+    if (err !=  TFM_HAL_SUCCESS) {
+        tfm_hal_its_aead_cleanup(&platform_ctx);
+        return err;
+    }
+
+
+    plat_err = nrf_cc3xx_platform_derived_key_encrypt(&platform_ctx,
+                                                      ciphertext,
+                                                      plaintext_size,
+                                                      plaintext);
+
+    tfm_hal_its_aead_cleanup(&platform_ctx);
+
+    if (plat_err != NRF_CC3XX_PLATFORM_SUCCESS) {
+        return TFM_HAL_ERROR_GENERIC;
+    }
+
+    return TFM_HAL_SUCCESS;
+}
+
+enum tfm_hal_status_t tfm_hal_its_aead_decrypt(
+                                        struct tfm_hal_its_auth_crypt_ctx *ctx,
+                                        uint8_t *ciphertext,
+                                        size_t ciphertext_size,
+                                        uint8_t *tag,
+                                        size_t tag_size,
+                                        uint8_t *plaintext,
+                                        size_t plaintext_size)
+{
+    nrf_cc3xx_platform_derived_key_ctx_t platform_ctx = {0};
+    enum tfm_hal_status_t err = TFM_HAL_ERROR_GENERIC;
+    int plat_err = NRF_CC3XX_PLATFORM_ERROR_INTERNAL;
+
+    if (!ctx_is_valid(ctx) || tag == NULL) {
+        return TFM_HAL_ERROR_INVALID_INPUT;
+    }
+
+    if (plaintext_size < ciphertext_size) {
+        return TFM_HAL_ERROR_INVALID_INPUT;
+    }
+
+    err = tfm_hal_its_aead_init(ctx,
+                                &platform_ctx,
+                                tag,
+                                tag_size);
+    if (err != TFM_HAL_SUCCESS) {
+        tfm_hal_its_aead_cleanup(&platform_ctx);
+        return err;
+    }
+
+
+    plat_err = nrf_cc3xx_platform_derived_key_decrypt(&platform_ctx,
+                                                      plaintext,
+                                                      ciphertext_size,
+                                                      ciphertext);
+    tfm_hal_its_aead_cleanup(&platform_ctx);
+
+    if (plat_err != NRF_CC3XX_PLATFORM_SUCCESS) {
+        return TFM_HAL_ERROR_GENERIC;
+    }
+
+    return TFM_HAL_SUCCESS;
+}
+
--- a/platform/ext/target/nordic_nrf/common/core/tfm_hal_its_encryption.h
+++ b/platform/ext/target/nordic_nrf/common/core/tfm_hal_its_encryption.h
@@ -1,0 +1,31 @@
+/*
+ * Copyright (c) 2023 Nordic Semiconductor ASA.
+ *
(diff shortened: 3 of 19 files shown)


platform/ext/target/nordic_nrf/common/core/tfm_hal_its_encryption.c
```diff
@@ -0,0 +1,288 @@
...
+            return TFM_HAL_ERROR_GENERIC;
+        }
+    }
+
+    memcpy(nonce, g_enc_nonce_seed, sizeof(g_enc_nonce_seed));
+    memcpy(nonce + sizeof(g_enc_nonce_seed),
+               &g_enc_counter,
+               sizeof(g_enc_counter));
+
+    g_enc_counter++;
```
SebastianBoe, line 75: The docs for tfm_hal_its_aead_generate_nonce highly recommend a PRNG.

So shouldn't we use that instead of
RNG + 0, RNG + 1, RNG + 2

i++

isn't a very good PRNG IMHO.

platform/ext/target/nordic_nrf/common/core/tfm_hal_its_encryption.c
```diff
@@ -0,0 +1,288 @@
...
+
+    return TFM_HAL_SUCCESS;
+}
```
SebastianBoe, line 78:
```suggestion
psa_status_t status = psa_generate_random(nonce, nonce_size);
return psa_status_to_tfm_hal_status(status);
```

Do we not have access to PSA APIs?

platform/include/tfm_hal_its.h
```diff
@@ -73,6 +78,121 @@ extern ARM_DRIVER_FLASH TFM_HAL_ITS_FLASH_DRIVER;
...
+ * \retval TFM_HAL_SUCCESS             The operation completed successfully
+ * \retval TFM_HAL_ERROR_INVALID_INPUT Invalid argument
+ */
+enum tfm_hal_status_t tfm_hal_its_aead_set_nonce(
+                                         struct tfm_hal_its_auth_crypt_ctx *ctx,
+                                         uint8_t *nonce,
+                                         size_t nonce_size);
+
+/**
+ * \brief Set the additional authenticated data for the AEAD operation.
```
SebastianBoe, line 132:
```suggestion
 * \brief Set the pointer to the additional authenticated data for the AEAD operation.
```

Let's make it more clear that no copying needs to be done here.

config/config_base.cmake
```diff
@@ -108,6 +108,9 @@ set(PS_ENCRYPTION                       ON          CACHE BOOL      "Enable encr
 set(PS_CRYPTO_AEAD_ALG                  PSA_ALG_GCM CACHE STRING    "The AEAD algorithm to use for authenticated encryption in Protected Storage")
```
SebastianBoe, line 108: Please cleanup the git history.

Parts of the second commit just undoes what happens in the first commit.

I don't see the point of that.

Also, there is no description in the second commit.

Git commits should contain why changes are being made.
What is happening should be self-evident from the code/comments.

secure_fw/partitions/internal_trusted_storage/tfm_its_req_mngr.c
```diff
@@ -130,6 +154,27 @@ static psa_status_t tfm_its_get_req(const psa_msg_t *msg)
...
+        /* When encryption is enabled the whole file needs to fit in the
+         * global buffer.
+         */
+        if(out_size > sizeof(asset_data)){
+            return PSA_ERROR_BUFFER_TOO_SMALL;
+        }
+    }
+#endif
+#endif
+
```
SebastianBoe, line 177:
```suggestion
   status = buffer_size_check(asset_info.client_id, out_size);
   if(status != PSA_SUCCESS) {
     return status;
   }
```

Let's re-use this code and minimize the amount of disturbance to the non-encrypted ITS.

SebastianBoe (changes requested):
changes requested

config/config_base.cmake
```diff
@@ -108,6 +108,9 @@ set(PS_ENCRYPTION                       ON          CACHE BOOL      "Enable encr
...
 
 set(TFM_PARTITION_INTERNAL_TRUSTED_STORAGE OFF      CACHE BOOL      "Enable Internal Trusted Storage partition")
+set(TFM_ITS_ENCRYPTED                   OFF         CACHE BOOL      "Enable authenticated encryption of ITS files using platform specific APIs")
+set(TFM_ITS_AUTH_TAG_LENGTH             "16"        CACHE STRING    "The size of the authentication tag used when authentication/encryption of ITS files is enabled ")
+set(TFM_ITS_ENC_NONCE_LENGTH            "16"        CACHE STRING    "The size of the nonce used when ITS file encryption is enabled")
```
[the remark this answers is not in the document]
mswarowsky: Me neither just took it to the previous construction, gonna change it.

config/config_base.cmake
mswarowsky: Yeah that is the plan after the review only count the first commit message

platform/include/tfm_hal_its.h
```diff
@@ -73,6 +78,121 @@ extern ARM_DRIVER_FLASH TFM_HAL_ITS_FLASH_DRIVER;
...
 tfm_hal_its_fs_info(struct tfm_hal_its_fs_info_t *fs_info);
 
+#ifdef TFM_ITS_ENCRYPTED
+
+/**
+ * \brief Generate an encryption nonce
+ *
+ * \details The nonce has to be unique for every encrypted file. A pseudorandom
+ *          number is highly reccomended but not enforced. Each platform need
+ *          to enforce their own strategy to differentiate the nonces.
```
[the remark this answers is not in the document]
mswarowsky: `highly recommended but not enforced` is actually a term I have read in combination with crypto schemes a couple of times. But at is has typos in it anyway gonna remove it :D

secure_fw/partitions/internal_trusted_storage/flash_fs/its_flash_fs.c
```diff
@@ -210,7 +210,7 @@ psa_status_t its_flash_fs_file_exist(struct its_flash_fs_ctx_t *fs_ctx,
 
 psa_status_t its_flash_fs_file_get_info(struct its_flash_fs_ctx_t *fs_ctx,
                                         const uint8_t *fid,
-                                        struct its_file_info_t *info)
+                                        struct its_flash_fs_file_info_t *info)
```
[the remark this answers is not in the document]
mswarowsky: Ok was done in a single commit in the upstream PR, and didn't get any complaints, but can fiddle that out and make a single commit. 
Even though I'm not sure I how to introduce that change later with a single PR upstream without having a broken state, as the type depends on the definition of TFM_ITS_ENCRYPTED

secure_fw/partitions/internal_trusted_storage/tfm_internal_trusted_storage.c
```diff
@@ -265,18 +282,50 @@ psa_status_t tfm_its_set(struct its_asset_info *asset_info,
...
+                        ITS_FLASH_FS_FLAG_CREATE | ITS_FLASH_FS_FLAG_TRUNCATE;
     }
 
+#ifdef TFM_ITS_ENCRYPTED
+#ifdef TFM_PARTITION_PROTECTED_STORAGE
+    if (client_id != TFM_SP_PS) {
+#else
+    {
+#endif /* TFM_PARTITION_PROTECTED_STORAGE */
+        /* When encryption is used, the while loop will execute only once
```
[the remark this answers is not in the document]
mswarowsky: Good catch, the code got updated from the old version, comment not :(

secure_fw/partitions/internal_trusted_storage/tfm_its_req_mngr.c
mswarowsky: good point, gonna put this in a sperate function

platform/ext/target/nordic_nrf/common/core/tfm_hal_its_encryption.c
mswarowsky: I gues @Vge0rge will have a better answer on it than me so leaving this open until he stated his design decisions

platform/ext/target/nordic_nrf/common/core/tfm_hal_its_encryption.c
Vge0rge: The main goal here was that we wanted to have unique IV for each file across resets. The fact that the IV is just IV++ is not an issue since it is still unique per reset and per file. The reason that we didn't use the RPNG for every call here was that is more costly and we didn't consider it adding security to the scheme. By using a unique PRNG per boot makes it nearly impossible to have the same IV for the same file twice across boots and that's enough. 

We can rephrase the documentation text to say that we highly suggest having a unique IV per file which is not the same across resets.

platform/ext/target/nordic_nrf/common/core/tfm_hal_its_encryption.c
Vge0rge: We don't have access to the PSA APIs, no :/ 

The reason for this is that the PSA crypto service stores the persistent keys in the ITS partition. So the ITS partition cannot go back to the PSA crypto service since this creates a cyclic dependency which is not accepted.

So we are forced to use platform specific APIs for this.

But even if we just implement a platform specific functionality which does an equivalent thing I think that it needs to be justified why we need to do the extra effort here and add the security advantages for this.

platform/ext/target/nordic_nrf/common/core/tfm_hal_its_encryption.h
```diff
@@ -0,0 +1,31 @@
+/*
+ * Copyright (c) 2022 Nordic Semiconductor ASA.
```
Vge0rge, line 2: We need to see if we need to update the Copyrights here.

secure_fw/partitions/internal_trusted_storage/flash_fs/its_flash_fs.c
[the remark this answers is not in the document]
Vge0rge: Gerrit supports chained PRs. So each commit here can be a separate PR in Gerrit which is chained so that it follows the commit history.

secure_fw/partitions/internal_trusted_storage/flash_fs/its_flash_fs.h
```diff
@@ -167,15 +167,23 @@ struct its_flash_fs_ops_t {
...
-    size_t size_current; /*!< The current size of the flash file data */
-    size_t size_max;     /*!< The maximum size of the flash file data in bytes.
-                          */
-    uint32_t flags;      /*!< Flags set when the file was created */
+struct its_flash_fs_file_info_t {
+    size_t size_current;  /*!< The current size of the file in bytes */
+    size_t size_max;      /*!< The maximum size of the file in bytes. */
+    uint32_t flags;       /*!< Flags set when the file was created */
+#ifdef TFM_ITS_ENCRYPTED
+    uint8_t add[ITS_FILE_ID_SIZE + 8];     /*!< Additional authenticated data */
```
Vge0rge, line 183: Could we improve the +8 here? Maybe add a define or something or an explanation what is this +8?

Vge0rge: I think that it is more clear to add a define where we say that the additional data are the ITS_FILE_SIZE_ID_SIZE + data size which is 4 bytes + flags which is 4 bytes as well.

secure_fw/partitions/internal_trusted_storage/tfm_its_req_mngr.c
```diff
@@ -66,6 +67,27 @@ static psa_status_t tfm_its_set_req(const psa_msg_t *msg)
     size_remaining = msg->in_size[1];
     offset = 0;
 
+#ifdef TFM_ITS_ENCRYPTED
+/* The buffer check is only need if we don't use the input buffer directly */
```
Vge0rge, line 71:
```suggestion
/* The buffer check is only needed if we don't use the input buffer directly */
```

secure_fw/partitions/internal_trusted_storage/tfm_its_req_mngr.c
```diff
@@ -66,6 +67,27 @@ static psa_status_t tfm_its_set_req(const psa_msg_t *msg)
     size_remaining = msg->in_size[1];
     offset = 0;
 
+#ifdef TFM_ITS_ENCRYPTED
```
Vge0rge, line 70: This is a difficult to read with the ifdefs. I think that it can be simplified  to be more readable so I have a suggestion. Use/update it if you think that it is more clear for you as well.

```suggestion
#if defined(TFM_ITS_ENCRYPTED) &&  (PSA_FRAMEWORK_HAS_MM_IOVEC != 1)

bool do_size_check = true;

#ifdef TFM_PARTITION_PROTECTED_STORAGE
if (asset_info.client_id != TFM_SP_PS) {
     do_size_check = false;
}
#endif 

        if(do_size_check && (size_remaining > sizeof(asset_data))){
            return PSA_ERROR_BUFFER_TOO_SMALL;
        }
 #endif
```

secure_fw/partitions/internal_trusted_storage/tfm_internal_trusted_storage.c
```diff
@@ -16,12 +16,28 @@
...
 #ifdef TFM_PARTITION_PROTECTED_STORAGE
 #include "ps_object_defs.h"
 #endif
 
+#ifdef TFM_ITS_ENCRYPTED
+
+/* Buffer to store the encrypted asset data before it is stored in the
+ * filesystem.
+ */
+static uint8_t enc_asset_data[ITS_UTILS_ALIGN(ITS_BUF_SIZE,
```
Vge0rge, line 32: Do we actually need two buffers? Do we use both of them at the same function call? I see that in get we do but I don't understand why we cannot just use one and skip the memcpy in the end?

platform/ext/target/nordic_nrf/common/core/tfm_hal_its_encryption.c
mswarowsky: Thanks for clarifying,  I rephrased the docs, but I like the hint with the PRN.

secure_fw/partitions/internal_trusted_storage/flash_fs/its_flash_fs.c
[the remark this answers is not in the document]
mswarowsky: Ohh ok, if they allow that, then we gonna "add" this as a new PR to the current open on

platform/ext/target/nordic_nrf/common/core/tfm_hal_its_encryption.c
mswarowsky: Ok then i would leave it @SebastianBoe you can resolve this conversation if you are fine with it

secure_fw/partitions/internal_trusted_storage/flash_fs/its_flash_fs.h
mswarowsky: Ahh I din't know myself what it is... 
Added two defines for it

secure_fw/partitions/internal_trusted_storage/tfm_internal_trusted_storage.c
mswarowsky: Needs investigation with Antonio, if in place decryption with `nrf_cc3xx_platform_derived_key_decrypt` and ChaCha20-Poly1305 is supported

secure_fw/partitions/internal_trusted_storage/tfm_its_req_mngr.c
mswarowsky: Ah I moved this out into its own function, hope that is easier to read

secure_fw/partitions/internal_trusted_storage/tfm_internal_trusted_storage.c
mswarowsky: Ok I changed it to an inline implementation, after we got the ok from arm

secure_fw/partitions/internal_trusted_storage/tfm_internal_trusted_storage.c
```diff
@@ -265,18 +282,47 @@ psa_status_t tfm_its_set(struct its_asset_info *asset_info,
...
-                                     size_to_write, offset, data_buf);
+                                     &g_file_info,
+                                     size_to_write, offset, buffer_pnt);
     if (status != PSA_SUCCESS) {
         return status;
     }
 
+    /* Do not create or truncate after the first iteration */
+    g_file_info.flags &= ~(ITS_FLASH_FS_FLAG_CREATE | ITS_FLASH_FS_FLAG_TRUNCATE);
+
```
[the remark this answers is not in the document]
mswarowsky: Good point changed it, as well for the get.

platform/ext/target/nordic_nrf/common/core/tfm_hal_its_encryption.c
SebastianBoe: This must be documented for future reference.

platform/include/tfm_hal_its.h
```diff
@@ -73,6 +78,120 @@ extern ARM_DRIVER_FLASH TFM_HAL_ITS_FLASH_DRIVER;
...
 tfm_hal_its_fs_info(struct tfm_hal_its_fs_info_t *fs_info);
 
+#ifdef TFM_ITS_ENCRYPTED
+
+/**
+ * \brief Generate an encryption nonce
+ *
+ * \details The nonce has to be unique for every encrypted file, even across
+ *          resets. Therefore it is recommended to base the nonce on a
+*           pseudorandom number.
```
SebastianBoe, line 88: We use the word therefore. But I don't see how the second sentence follows from the first.

It's not clear to me what "base the nonce on" means in the context of this sentence either.

platform/ext/target/nordic_nrf/common/core/tfm_hal_its_encryption.c
```diff
@@ -0,0 +1,288 @@
...
+    }
+
+    if(nonce_size < sizeof(g_enc_nonce_seed) + sizeof(g_enc_counter)){
+        return TFM_HAL_ERROR_INVALID_INPUT;
+    }
+
+    if (g_enc_counter == 0) {
+        if (sizeof(g_enc_nonce_seed) != 8) {
+            return TFM_HAL_ERROR_GENERIC;
+        }
```
SebastianBoe, line 62: So TFM_ITS_ENC_NONCE_LENGTH is configurable but for our implementation it must be 12.

We should make this a build time failure, not a runtime failure.

platform/ext/target/nordic_nrf/common/core/tfm_hal_its_encryption.c
```diff
@@ -0,0 +1,288 @@
...
+ */
+static uint32_t g_enc_counter;
+
+/* The global nonce seed which is fetched once in every boot. The seed is used
+ * as part of the nonce and allows the platforms to diversify their nonces
+ * across resets. Note that the way that this seed is generated is platform
+ * specific, so the diversification is optional.
+ */
+static uint8_t g_enc_nonce_seed[TFM_ITS_ENC_NONCE_LENGTH -
+                                sizeof(g_enc_counter)];
```
SebastianBoe, line 43:
```suggestion
static uint8_t g_enc_nonce_seed[8];
```

BUILD_ASSERT(TFM_ITS_ENC_NONCE_LENGTH == 12, "This implementation only supports a nonce of size 12");

platform/ext/target/nordic_nrf/common/core/tfm_hal_its_encryption.c
Vge0rge: TF-M doesn't allow cyclic dependencies.

platform/ext/target/nordic_nrf/common/core/tfm_hal_its_encryption.c
SebastianBoe: BUILD_ASSERT(TFM_ITS_ENC_NONCE_LENGTH == 12, "This implementation only supports a nonce of size 12");

platform/ext/target/nordic_nrf/common/core/tfm_hal_its_encryption.c
Vge0rge: Or to be more specific the PSA Firmware Framework doesn't allow circular dependencies. You can search for "circular" in the framework.

platform/include/tfm_hal_its.h
mswarowsky: Ok, any better suggestions? Then we just use that one ;)

SebastianBoe: "The nonce must be unique for every encrypted file, even across resets."

AFAICT it is implementation defined whether or not it is a good idea to use a pseudorandom number to implement this, or a non-volatile monotonic counter.

This is based on the nonce not being a secret, which I don't know if it is or not.

platform/ext/target/nordic_nrf/common/core/tfm_hal_its_encryption.c
mswarowsky: added

platform/ext/target/nordic_nrf/common/core/tfm_hal_its_encryption.c
mswarowsky: I added a short comment

platform/ext/target/nordic_nrf/common/core/tfm_hal_its_encryption.c
```diff
@@ -0,0 +1,296 @@
...
+
+/* As TFM_ITS_ENC_NONCE_LENGTH is configurable but this implementation expects
+ * the seed to be 8 byes the nonce length should always be 12.
```
SebastianBoe, line 46:
```suggestion
 * the seed to be 8 bytes and the nonce length to always be 12.
```

platform/ext/target/nordic_nrf/common/core/tfm_hal_its_encryption.c
```diff
@@ -0,0 +1,296 @@
...
+    }
+
+    if(nonce_size < sizeof(g_enc_nonce_seed) + sizeof(g_enc_counter)){
+        return TFM_HAL_ERROR_INVALID_INPUT;
+    }
+
+    if (g_enc_counter == 0) {
+        if (sizeof(g_enc_nonce_seed) != 8) {
+            return TFM_HAL_ERROR_GENERIC;
+        }
```
SebastianBoe, line 69:
```suggestion
```

Why do we have this code?

Note that it will always be compiled out due to sizeof being evaluated at compile-time.

secure_fw/partitions/internal_trusted_storage/its_crypto_interface.c
```diff
@@ -0,0 +1,186 @@
...
+                                fid_size,
+                                finfo->flags,
+                                file_size);
+    if (err != TFM_HAL_SUCCESS) {
+        return tfm_hal_to_psa_error(err);
+    }
+
+    if (is_encrypt) {
+        err = tfm_hal_its_aead_generate_nonce(finfo->nonce,
+                                              sizeof(finfo->nonce));
```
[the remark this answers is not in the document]
mswarowsky: Again @Vge0rge eventually can give an explanation why this approach was chosen. 
Without looking into it I wonder what happens if all ITS files get deleted. How would we then prevent starting from 0 again?

Could also be that I don't understand the suggestions correctly.

platform/include/tfm_hal_its.h
mswarowsky: Ok, I thought it is nice to give people a hint, of what would be a good approach instead of leaving the fully on their own, but no strong will on this removed the second sentence.

platform/ext/target/nordic_nrf/common/core/tfm_hal_its_encryption.c
```diff
@@ -0,0 +1,296 @@
...
+        /* The cc3xx driver is used directly as the PSA APIs depends on ITS, so
+         * using PSA APIs would result in circular dependencies */
```
SebastianBoe, line 71: This comment should be in the file-level documentation as it applies to the design of all the functions in this C file and not just this specific function call.

secure_fw/partitions/internal_trusted_storage/its_crypto_interface.c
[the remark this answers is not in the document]
SebastianBoe: That is a good point.

I had another idea.

What about using the file ID as the nonce?

I am unable to think of any problems with this approach.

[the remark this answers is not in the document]
SebastianBoe: Hmm, actually, I wonder if this might break the unique requirement somehow ...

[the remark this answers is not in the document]
SebastianBoe: Yeah, I think it has the same problem as you identified for a montonic counter. When you wipe the filesystem you can encrypt a different plaintext with the same nonce.

platform/ext/target/nordic_nrf/common/core/tfm_hal_its_encryption.c
```diff
@@ -0,0 +1,296 @@
...
+        if (err != 0) {
+            return TFM_HAL_ERROR_GENERIC;
+        }
+    }
+
+    memcpy(nonce, g_enc_nonce_seed, sizeof(g_enc_nonce_seed));
+    memcpy(nonce + sizeof(g_enc_nonce_seed),
+               &g_enc_counter,
+               sizeof(g_enc_counter));
+
```
SebastianBoe, line 82:
```
/* To avoid wrap-around of the g_enc_counter and subsequent re-use of the nonce
 * we check for a counter value so large that it could only be nefariously produced.
 * /
if(g_enc_counter > 100 * 1000 * 1000) {
   return TFM_HAL_ERROR_GENERIC;
}
```

An attacker can construct input such that tfm_hal_its_aead_generate_nonce succeeds, but a later stage to fails, and then increment the counter until we re-use the nonce.

SebastianBoe: I suppose this could should be higher up, as a part of input validation.

secure_fw/partitions/internal_trusted_storage/its_crypto_interface.c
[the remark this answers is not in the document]
SebastianBoe: According to RFC 5116 g_enc_nonce_seed must be fixed across resets and unique per device. Presumably to avoid the birthday collision problem.

"The Fixed field MUST remain constant for all nonces that are generated for a given encryption device."

https://www.rfc-editor.org/rfc/rfc5116#section-3.1

Our current approach appears to be violating a MUST statement of the RFC and therefore be unsecure unless I am missing something.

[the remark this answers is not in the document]
SebastianBoe: I double-checked this, and it's not the same scheme, our counter isn't monotonic.

Do we have a trustworthy source, like an RFC, that says our approach to the the nonce is secure?

SebastianBoe (approved):

SebastianBoe (changes requested):
changes requested

secure_fw/partitions/internal_trusted_storage/its_crypto_interface.c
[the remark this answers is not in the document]
mswarowsky: After offline discussion added a disclaimer at the  beginning of `tfm_hal_its_aead_generate_nonce`

Anything else will be handled in the updated documentation afterwards

platform/ext/target/nordic_nrf/common/core/tfm_hal_its_encryption.c
Vge0rge: This seems unneeded indeed.  We can remove it.

secure_fw/partitions/internal_trusted_storage/tfm_its_req_mngr.c
```diff
@@ -32,6 +33,26 @@ static uint8_t __ALIGNED(4) asset_data[ITS_UTILS_ALIGN(ITS_BUF_SIZE,
                                           ITS_FLASH_MAX_ALIGNMENT)];
 #endif
 
+
+static psa_status_t buffer_size_check(int32_t client_id, size_t buffer_size) {
```
maciejbaczmanski, line 37: Hi @mswarowsky .
I've noticed that when building samples for `nrf5340dk_nrf5340_cpuapp_ns` for openthread cli I'm getting a warning: 
```
tfm_its_req_mngr.c:37:21: warning: 'buffer_size_check' defined but not used [-Wunused-function]
   37 | static psa_status_t buffer_size_check(int32_t client_id, size_t buffer_size) {
      |                     ^~~~~~~~~~~~~~~~~
```
I think adding an `#ifdef TFM_ITS_ENCRYPTED` guard here is necessary too.

joerchan: @maciejbaczmanski Already created a PR here: https://github.com/nrfconnect/sdk-trusted-firmware-m/pull/88

maciejbaczmanski: Oh, didn't see that. Thanks!

PR #97, opened 2023-07-05
Target cfg

SebastianBoe:


platform/ext/target/nordic_nrf/common/core/native_drivers/spu.c
```diff
@@ -126,27 +126,28 @@ void spu_regions_reset_unlocked_secure(void)
     for (size_t i = 0; i < NUM_FLASH_SECURE_ATTRIBUTION_REGIONS ; i++) {
```
joerchan, line 126: For spu_peripheral_config_secure/non_secure.

If we add:
SPU_SECURE_DMA_SECURE
SPU_SECURE_DMA_NONSECURE

Then we can add reduce these two functions to one spu_config_peripheral, add secure_attr and secure_dma as argument, and  rename periph_lock to lock_conf.

Not sure if worth it though, maybe just make it a common function to be used by those two functions instead?

platform/ext/target/nordic_nrf/common/core/native_drivers/spu.h
```diff
@@ -23,6 +23,10 @@
 
 #include <hal/nrf_spu.h>
 
+#define SPU_LOCK_CONF_LOCKED true
```
joerchan, line 26: dppi and gpio can also use LOCK_CONF defines.

platform/ext/target/nordic_nrf/common/core/target_cfg.c
```diff
@@ -703,29 +703,38 @@ enum tfm_plat_err_t spu_init_cfg(void)
      */
     spu_regions_reset_unlocked_secure();
 
+	uint32_t perm;
```
joerchan, line 706: Mixed indentation style.

platform/ext/target/nordic_nrf/common/core/native_drivers/spu.c
SebastianBoe: discussed offline. Will treat the peripheral API as out-of-scope.

PR #98, opened 2023-07-06
TF-M startup file improvements

SebastianBoe:
TF-M startup file improvements.

See commit messages for details.

noup because toup is not allowed.
Will push to upstream after review and merging here.


joerchan:
> noup because toup is not allowed.

TF-M has exception, so you could toup it if you want.

platform/ext/target/nordic_nrf/common/core/CMakeLists.txt
```diff
@@ -71,6 +75,7 @@ target_sources(platform_s
         cmsis_drivers/Driver_Flash.c
         ${HAL_NORDIC_PATH}/nrfx/drivers/src/nrfx_nvmc.c
         nrfx_glue.c
+        startup.c
```
joerchan, line 78: You're missing BL2?

platform/ext/target/nordic_nrf/common/core/startup.c
```diff
@@ -0,0 +1,68 @@
+/*
+ * Copyright (c) 2022 Arm Limited. All rights reserved.
```
joerchan, line 2: Please put the nordic license here, same as the header file.

platform/ext/target/nordic_nrf/common/nrf5340/gcc/startup_nrf5340.c
```diff
@@ -25,6 +25,10 @@
 #include "hw_init.h"
 #include "startup.h"
 
+__NO_RETURN __attribute__((naked)) void default_tfm_IRQHandler(void) {
```
joerchan, line 29: Why not in the common startup.c file?

platform/ext/target/nordic_nrf/common/core/startup.c
```diff
@@ -25,6 +25,27 @@
...
+{
+	/*
+	 * This file is used by the TF-M build system. Usually in the
+	 * secure domain (tfm_s.elf), but also for testing purposes in the
+	 * non-secure domain (tfm_ns.elf).
+	 *
+	 * When used in the non-secure domain we do not have
+	 * tfm_core_panic so we loop forever instead.
+	 */
+#ifdef DOMAIN_NS
```
joerchan, line 42: I suspect you need this for BL2 as well? BL2 will not have tfm_core_panic and will have compilation error?

platform/ext/target/nordic_nrf/common/core/startup.h
```diff
@@ -28,6 +28,8 @@ __NO_RETURN void __attribute__((weak, alias("default_tfm_IRQHandler"))) handler_
 
 __NO_RETURN void Reset_Handler(void);
 
+void SPU_IRQHandler(void);
```
joerchan, line 31: This build system bug has happened before. For the other types of fault.
Is it a regression since then for SPU possibly?

joerchan (changes requested):

platform/ext/target/nordic_nrf/common/core/startup.h
joerchan: FYI: https://github.com/nrfconnect/sdk-trusted-firmware-m/commit/4e1d40b1e108ce9c0f7fa610bf69f65c0fd28305

platform/ext/target/nordic_nrf/common/nrf5340/gcc/startup_nrf5340.c
SebastianBoe: Because it did not compile when I did so.

I tried different things but I was not able to figure out why I could not do this.

platform/ext/target/nordic_nrf/common/core/startup.c
SebastianBoe: Is that for vanilla TF-M?

NCS TF-M does not use BL2 right?

platform/ext/target/nordic_nrf/common/core/startup.h
SebastianBoe: I would assume it is a regression since then yes :)

I could try adding the source to tfm_s.

But IMHO it is more obvious where the SPU_IRQ comes from the way I did than to have a default weak entry that is dead code.

platform/ext/target/nordic_nrf/common/core/CMakeLists.txt
joerchan, line 78: Also, startup file should be in `tfm_s` not `platform_s`

platform/ext/target/nordic_nrf/common/core/startup.h
joerchan: Sure, I don't object to this change, but I think it could be extended to include the rest of the faults as well, so we don't end up with the same bug again.

joerchan: It makes more sense to hav ethe startup file in tfm_s, since that is the lib where startup_nrf5340.c is for example.

platform/ext/target/nordic_nrf/common/core/startup.c
SebastianBoe: Maybe BL2 is used for "our" MCUBoot integration?

joerchan: I was reviewing this patch in the context that you would put this upstream, so then you need the BL2 support.

Vge0rge (approved):

platform/ext/target/nordic_nrf/common/core/startup.h
SebastianBoe: startup.c is now added to tfm_s

platform/ext/target/nordic_nrf/common/core/CMakeLists.txt
```diff
@@ -162,6 +171,7 @@ if(BL2)
         PRIVATE
             cmsis_drivers/Driver_Flash.c
             cmsis_drivers/Driver_USART.c
+            $<$<C_COMPILER_ID:GNU>:${CMAKE_CURRENT_SOURCE_DIR}/startup.c>
```
joerchan, line 174: This file should be added to `bl` library (same as the board specific startup files).

SebastianBoe: fixed

joerchan (approved):

PR #99, opened 2023-07-17
startup_file_improvements: [nrf noup] platform: nordic_nrf: Bugfix in startup file

SebastianBoe:
Fix a bug in the startup file where the ifdef was incorrectly negated.

Also add declarations for when the default irq handler is not used.


frkv (approved):
LGTM

joerchan (approved):

PR #103, opened 2023-08-11
[nrf noup] platform: nrf_nordic: Add temp fix for for secure FPU on 9161

mswarowsky:
The MDK for nRF9120 used in the nRF9161 target doesn't define the Secure FPU as it doesn't exist, but for other platforms like the 9160 it has a dummy define, with an UNUSED field in the type.
The long plan is to get this fixed in the MDK but until then, to make the nrfxlib 3.1.0 update possible this tempfix is applied.

 Ref: NCSDK-23046

Change-Id: I44042ee9aada99c59a5930440306bb6c40ae4880


frkv (approved):
LGTM! I hear it is going to be added here: https://github.com/nrfconnect/sdk-nrf/pull/11800

PR #112, opened 2023-11-07
Fix psa want missing

joerchan:
1.
Fix regression since target_cfg.c was refactored to a common file
for nrf91 and nrf53.
The I2S and PDM peripheral are not configured as non-secure peripherals.

2.
Disable the cipher crypto module in small, medium and mediuam-arotless
profile. There is no algorithm for this module enabled in the mbedcrypto
configuration header for these profiles.
Add missing PSA defined algorithms and keys checks.

The checks only covered supported algorithms in mbedtls.
However mbedtls supports accelerated PSA crypto support through the
psa crypto driver wrappers, which can support additional algorithms
and key types.

### DIFF
--- a/secure_fw/partitions/crypto/crypto_check_config.h
+++ b/secure_fw/partitions/crypto/crypto_check_config.h
@@ -9,62 +9,113 @@
 
 #include "config_tfm.h"
 
-#if CRYPTO_RNG_MODULE_ENABLED  && \
-    (!defined(MBEDTLS_CTR_DRBG_C) &&            \
-     !defined(MBEDTLS_HMAC_DRBG_C) &&           \
-     !defined(MBEDTLS_PSA_CRYPTO_EXTERNAL_RNG))
-#error "CRYPTO_RNG_MODULE_ENABLED enables, but not all prerequisites (missing RNG)!"
+/* RNG module config can't be adjusted on PSA_WANT_* requirements. */
+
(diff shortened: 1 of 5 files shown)


vili-nordic (approved):

mswarowsky (approved):

secure_fw/partitions/crypto/crypto_check_config.h
```diff
@@ -9,62 +9,113 @@
...
 #include "config_tfm.h"
 
-#if CRYPTO_RNG_MODULE_ENABLED  && \
-    (!defined(MBEDTLS_CTR_DRBG_C) &&            \
-     !defined(MBEDTLS_HMAC_DRBG_C) &&           \
-     !defined(MBEDTLS_PSA_CRYPTO_EXTERNAL_RNG))
-#error "CRYPTO_RNG_MODULE_ENABLED enables, but not all prerequisites (missing RNG)!"
+/* RNG module config can't be adjusted on PSA_WANT_* requirements
+ * because the PSA Crypto subsystem guarantees on booting that a
+ * source of entropy with a configured deterministic random number
```
Vge0rge, line 14: I would propose to change the wording so that it doesn't say deterministic random number generator. 
Probably better to replace it with something like: cryptographically secure pseudorandom number generator

After all it is definitely not  deterministic.

joerchan: I just removed all of that. Not important information.

PR #115, opened 2023-11-22
[nrf noup] platform: nordic_nrf: Fix NS build with floating point ena…

joerchan:
…bled

Fix TF-M building of the NS application with floating point enabled. This is needed in order to run the TF-M regression tests.

NOUP:
This does not have an upstream PR as this problem does not exits in the main branch after the TF-M split NS build feature merge. This commit should therefore just be reverted on next TF-M sync.


frkv (approved):
LGTM

magnev (approved):

PR #124, opened 2024-02-29
[nrf noup] crypto: Add PAKE support

mswarowsky:
This is noup commit as upstream TF-M relies on the mbed TLS PSA Core that does not support the PAKE API's according to 1.2 at the moment. Once this exists then this can be up streamed, or removed if TF-M adds it themself.

Added PAKE API support according to the PSA crypto spec 1.2

Ref: NCSDK-22416

### DIFF
--- a/secure_fw/partitions/crypto/crypto_check_config.h
+++ b/secure_fw/partitions/crypto/crypto_check_config.h
@@ -118,4 +118,14 @@
 #error "CRYPTO_KEY_DERIVATION_MODULE_ENABLED enabled, but not all prerequisites (missing key derivation algorithms)!"
 #endif
 
+#if CRYPTO_PAKE_MODULE_ENABLED && \
+    (!defined(PSA_WANT_ALG_JPAKE) && \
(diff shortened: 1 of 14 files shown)


secure_fw/partitions/crypto/crypto_check_config.h
```diff
@@ -118,4 +118,15 @@
 #error "CRYPTO_KEY_DERIVATION_MODULE_ENABLED enabled, but not all prerequisites (missing key derivation algorithms)!"
 #endif
 
+/* TODO check what is needed here */
```
SebastianBoe, line 121: did you forget this TODO or would you like to keep it? :)

frkv (approved):
LGTM (minus TODO)

SebastianBoe (approved):

secure_fw/partitions/crypto/crypto_check_config.h
mswarowsky: Good catch, removed the TODO

vili-nordic (approved):

magnev (approved):

PR #125, opened 2024-03-01
platform: nordic_nrf: Add support for 54l

SebastianBoe:
Change-Id: I650ab5309cdea7f3508be7cfb124e0032c13d777


mswarowsky (approved):
Maybe we should stick with the commit tag scheme, but not very important for me

vili-nordic (approved):

SeppoTakalo (changes requested):
I don't think we can accept !fixup commits. That may ruin someones rebase, if they have `autofixup` enabled.


Remove the tag.

SebastianBoe:
@jfischer-no : We are re-writing the history before release right?

SebastianBoe:
@SeppoTakalo : I don't think that scenario would happen before the history was rewritten, but I have updated accordingly.

SeppoTakalo (approved):

PR #128, opened 2024-03-04
cmake: tools: Dont add a custom command for running the manifest

SebastianBoe:
In the TF-M build scripts we run the manifest tool twice, first from CMake and then from ninja.

It is bad practice to configure CMake projects like this. Instead, if configuration from CMake is necessary, one should configure from CMake only, and then re-run CMake when necessary, not just the command.

This organization has been causing problems for our users as they have been required to rebuild TF-M twice.

This is due to this scenario playing out:

CMake generates config_impl.cmake by invoking the manifest tool at Configure time.

CMake generates build.ninja.

Ninja generates config_impl.cmake by invoking the manifest tool at build time.

When the user then invokes ninja a second time config_impl.cmake will be newer than build.ninja. But CMake is supposed to be includ'ing config_impl.cmake, so build.ninja is now considered out-of-date wrt. config_impl.cmake.

ninja therefore invokes CMake again, and then ninja afterwards.

Change-Id: Icef588479d27fa3a172b40b09eacad417922fba5

### DIFF
--- a/tools/CMakeLists.txt
+++ b/tools/CMakeLists.txt
@@ -159,12 +159,17 @@ set(MANIFEST_COMMAND
     -o ${CMAKE_BINARY_DIR}/generated
     ${PARSE_MANIFEST_QUIET_FLAG})
 
+set(NO_BUILD_CMD_FOR_MANIFEST 1)
+
+if(NO_BUILD_CMD_FOR_MANIFEST)
+else()
 add_custom_command(
     OUTPUT ${CMAKE_BINARY_DIR}/generated
     COMMAND ${MANIFEST_COMMAND}
     DEPENDS ${MANIFEST_LISTS} ${GENERATED_FILE_LISTS}
             ${MANIFEST_FILES} ${TEMPLATE_FILES}
 )
+endif()
 
 add_custom_target(
     manifest_tool


mswarowsky (approved):
Nice work

Vge0rge (approved):

tools/CMakeLists.txt
```diff
@@ -159,12 +159,17 @@ set(MANIFEST_COMMAND
     -o ${CMAKE_BINARY_DIR}/generated
     ${PARSE_MANIFEST_QUIET_FLAG})
 
+set(NO_BUILD_CMD_FOR_MANIFEST 1)
```
balaji-nordic, line 162: If I understand correctly, the intention here is to set/unset this variable in the TF-M build script whenever we do not want to run the manifest command. Can you point me to how this variable will be used in the build script?

balaji-nordic (approved):
The change works (tested locally on ncs 2.6.0 release candidate). Just have a few suggestions. You may or may not address them

SeppoTakalo:
This feels hack.

If you aim to do
```
set(A 1)
if (NOT A) {
```

There seem to be no way to run the block anymore. Therefore, why don't you just remove the block?
What is the purpose of it?

SebastianBoe:
It's temporary code.

After some more user-testing (build system dependency issues are difficult to test without user-testing)
the set(A 1) line will be moved out of the TF-M project and into the Zephyr Cmake build system.

We don't immediately start the upstreaming process because it is expensive to do multiple upstreaming iterations,
I would prefer to wait and see if this code works as intended without side-effects.

PR #131, opened 2024-03-08
nRF54L15: Add initial TF-M support

Vge0rge:


platform/ext/target/nordic_nrf/common/core/target_cfg.c
```diff
@@ -46,6 +46,59 @@
 #define PIN_XL1 0
 #define PIN_XL2 1
 
+/* Use the SPU_PERIPH_PERM_SECUREMAPPING_NonSecure symbol to determine
+ * if we have and old or new SPU.
+ */
+#ifdef SPU_PERIPH_PERM_SECUREMAPPING_NonSecure
+/* TODO: Remove this temporary fix when the patch from NRFX-3570 is available.
```
mswarowsky, line 53: I know how bad our memory is so maybe add a Jira to this TODO :)

mswarowsky (approved):
On small nit for our memory, but else fine

SebastianBoe (approved):

platform/ext/target/nordic_nrf/common/core/target_cfg.c
SebastianBoe: we'll get a multiple definition error when NRFX is updated so it shouldn't be necessary to have a reminder :)

PR #139, opened 2024-04-19
[nrf noup] Add MPC and SPC error reporting for nrf54l

vili-nordic:
Adds handling of MPC and SPC errors.

### DIFF
--- a/platform/ext/target/nordic_nrf/common/core/nrf_exception_info.c
+++ b/platform/ext/target/nordic_nrf/common/core/nrf_exception_info.c
@@ -11,34 +11,60 @@
 
 static struct nrf_exception_info nrf_exc_info;
 
-static void spu_dump_context(struct nrf_exception_info *ctx)
+static void dump_exception_info(struct nrf_exception_info *ctx)
 {
-    SPMLOG_ERRMSG("Platform Exception: SPU Fault\r\n");
+    SPMLOG_ERRMSG("Platform Exception:\r\n");
 
-#ifdef NRF_SPU
     /* Report which type of violation occured */
     if (ctx->events & SPU_EVENT_RAMACCERR) {
-        SPMLOG_DBGMSG("  RAMACCERR\r\n");
+        SPMLOG_DBGMSG("  SPU.RAMACCERR\r\n");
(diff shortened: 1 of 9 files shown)


SebastianBoe (approved):

mswarowsky (approved):
This is gonna be so nice to have the address in the report.

Could you add maybe 
`fixup! [nrf noup] platform: nordic_nrf: Add support for 54l`
as first line of the commit description so this will get squashed no the next upmerge

platform/ext/target/nordic_nrf/common/core/nrf_exception_info.c
```diff
@@ -11,34 +11,60 @@
 
 static struct nrf_exception_info nrf_exc_info;
 
-static void spu_dump_context(struct nrf_exception_info *ctx)
+static void dump_exception_info(struct nrf_exception_info *ctx)
 {
     SPMLOG_ERRMSG("Platform Exception: SPU Fault\r\n");
```
Vge0rge, line 16: Very nit but this should write something like more generic now that MPC maybe caused the fault?

platform/ext/target/nordic_nrf/common/core/nrf_exception_info.c
```diff
@@ -11,34 +11,60 @@
...
+        SPMLOG_DBGMSGVAL("    Execute:      ", (ctx->memaccerr.info >> 14) & 1);
+        SPMLOG_DBGMSGVAL("    Secure:       ", (ctx->memaccerr.info >> 15) & 1);
+        SPMLOG_DBGMSGVAL("    Error source: ", (ctx->memaccerr.info >> 16) & 1);
+    }
+#endif
 }
 
 void nrf_exception_info_store_context(void)
 {
     nrf_exc_info.events = spu_events_get();
```
Vge0rge, line 50: Does that not need to be inside the ifdef?

vili-nordic: No, this is platform independent, the code in the ifdefs are to get extra information on supported platforms.

PR #143, opened 2024-05-02
 Enable Protected Storage encryption

mswarowsky:

PR #144, opened 2024-05-08
[nrf noup] platform: nordic_nrf:  config of UART instances

mswarowsky:
fixup! [nrf noup] platform: nordic_nrf: Add support for 54l

This allows select which UART instance is used for TF-M

Ref: NCSDK-25009


mswarowsky:
Got merged with #143 143

PR #148, opened 2024-05-21
[nrf noup] cmake: Fix TFM psa_crypto_config linking error

rlubos:
Fix linking errors with psa_crypto_config observed in TFM test applications.

To be reverted during the next TFM upmerge, as this isolated change is already part of a larger commit upstream.


frkv (approved):
LGTM (Hoping testing agrees)

mswarowsky (approved):

PR #151, opened 2024-05-23
[nrf fromtree] Platform: nordic_nrf: Don't configure NRF_VMC as non-s…

nordic-mare:
…ecure

Dont configure the volatile memory controller as a non-secure peripheral

Change-Id: I2489defaf6deb89beba7447ba079ea3e5afebca5

### DIFF
--- a/platform/ext/target/nordic_nrf/common/core/target_cfg.c
+++ b/platform/ext/target/nordic_nrf/common/core/target_cfg.c
@@ -1285,7 +1285,6 @@ static const uint32_t target_peripherals[] = {
 #ifdef NRF_P1
     NRF_P1_S_BASE,
 #endif
-    NRF_VMC_S_BASE,
 };
 
     for (int i = 0; i < ARRAY_SIZE(target_peripherals); i++) {


mswarowsky:
Maybe add `(cherry picked from commit c670a6af1f0a3d7d6389e8879e8de17c1bd442fe)` to the commit message.

Also, we need a manifest PR to ensure we don't introduce a regression with it, which will bock other PR's

mswarowsky (approved):

vili-nordic (approved):

Vge0rge (approved):

SebastianBoe:
Hi,

we also need to update the config system.

Or else we will have a mismatch between what the config system is describing, and what the implemenation is doing.

i.e. like this:

~/ncs/zephyr/samples/hello_world$ west build -p -b nrf9161dk/nrf9161/ns

ack _VMC_SECURE
build/hello_world/zephyr/.config
1044:# CONFIG_NRF_VMC_SECURE is not set


In Kconfig we have encoded that CONFIG_NRF_VMC_SECURE=n, but in reality it is CONFIG_NRF_VMC_SECURE=y.

PR #153, opened 2024-05-27
[noup] nordic_nrf: 54l Add UARTE defines to peripherals config

mswarowsky:
This is needed to make the tfm_regression tests run


platform/ext/target/nordic_nrf/common/nrf54l15/config.cmake
```diff
@@ -8,7 +8,7 @@
 
 include(${PLATFORM_PATH}/common/core/config.cmake)
 
-set(SECURE_UART1                        ON         CACHE BOOL      "Enable secure UART1")
+set(SECURE_UART30                       ON         CACHE BOOL      "Enable secure UART1")
```
SebastianBoe, line 11:
```suggestion
set(SECURE_UART30                       ON         CACHE BOOL      "Enable secure UART")
```

PR #158, opened 2024-06-21
[nrf noup] Fix version warning

vili-nordic:
Version check depends on upstream's tagging scheme which differs from NCS's


frkv (approved):
LGTM

endre-nordic (approved):

PR #161, opened 2024-07-25
[nrf noup] platform: nordic_nrf: Add support for 54l

Vge0rge:
fixup! [nrf noup] platform: nordic_nrf: Add support for 54l

Clear the MPC events before you enable the MPC interrupt.

### DIFF
--- a/platform/ext/target/nordic_nrf/common/core/faults.c
+++ b/platform/ext/target/nordic_nrf/common/core/faults.c
@@ -86,8 +86,10 @@ __attribute__((naked)) void SPU30_IRQHandler(void)
 #endif
 
 #ifdef NRF_MPC00
-void MPC_Handler(void)
+__attribute__((naked)) void MPC_Handler(void)
 {
+    EXCEPTION_INFO();
+
 #ifdef TFM_EXCEPTION_INFO_DUMP
     nrf_exception_info_store_context();
 #endif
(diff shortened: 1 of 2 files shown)


vili-nordic (approved):

vili-nordic (approved):

tomi-font (commented):
How does this `Add support for 54l`? And why as a `noup`?

Vge0rge:
> How does this `Add support for 54l`? And why as a `noup`?

Ouuupsi, I updated the description of the commit but not the title, sorry. I will update the title as well. 

This is a fixup to the commit that we add support for nrf54L which is not upstreamed yet. Since the commit which adds support for it is not upstreamed it makes sense to have this as a fixup commit so that in the next upmerge we can combine them. The original commit is a noup because we don't have a specific timeline on when we will push it in the upstream repo and the toup recipe is not allowed anymore.

tomi-font:
> > How does this `Add support for 54l`? And why as a `noup`?
> 
> Ouuupsi, I updated the description of the commit but not the title, sorry. I will update the title as well.
> 
> This is a fixup to the commit that we add support for nrf54L which is not upstreamed yet. Since the commit which adds support for it is not upstreamed it makes sense to have this as a fixup commit so that in the next upmerge we can combine them. The original commit is a noup because we don't have a specific timeline on when we will push it in the upstream repo and the toup recipe is not allowed anymore.

How about instead pushing the change upstream and turning this into a `fromlist` even if it doesn't get reviewed right away upstream to still keep track of it?

Vge0rge:
> > > How does this `Add support for 54l`? And why as a `noup`?
> > 
> > 
> > Ouuupsi, I updated the description of the commit but not the title, sorry. I will update the title as well.
> > This is a fixup to the commit that we add support for nrf54L which is not upstreamed yet. Since the commit which adds support for it is not upstreamed it makes sense to have this as a fixup commit so that in the next upmerge we can combine them. The original commit is a noup because we don't have a specific timeline on when we will push it in the upstream repo and the toup recipe is not allowed anymore.
> 
> How about instead pushing the change upstream and turning this into a `fromlist` even if it doesn't get reviewed right away upstream to still keep track of it?

It's not in scope for this task I think. This change by itself cannot be pushed upstream, we will have to do some cleanup in this repo and then push full support for 54L15 upstream. That's a much bigger task so I don't want to do this as part of this relatively small change.

platform/ext/target/nordic_nrf/common/core/faults.c
```diff
@@ -86,8 +86,10 @@ __attribute__((naked)) void SPU30_IRQHandler(void)
 #endif
 
 #ifdef NRF_MPC00
-void MPC_Handler(void)
+__attribute__((naked)) void MPC_Handler(void)
```
tomi-font, line 89: Seems like this function shouldn't be `naked`: https://gcc.gnu.org/onlinedocs/gcc/ARM-Function-Attributes.html#index-naked-function-attribute_002c-ARM

platform/ext/target/nordic_nrf/common/core/target_cfg.c
```diff
@@ -896,6 +896,7 @@ enum tfm_plat_err_t nvic_interrupt_enable(void)
 #ifdef MPC_PRESENT
     /* MPC interrupt enabling */
     mpc_enable_interrupts();
+    mpc_clear_events();
```
tomi-font, line 899: Shouldn't you swap those two lines?

tomi-font:
> It's not in scope for this task I think. This change by itself cannot be pushed upstream, we will have to do some cleanup in this repo and then push full support for 54L15 upstream. That's a much bigger task so I don't want to do this as part of this relatively small change.

Alright.

platform/ext/target/nordic_nrf/common/core/target_cfg.c
Vge0rge: done

platform/ext/target/nordic_nrf/common/core/target_cfg.c
```diff
@@ -894,6 +894,7 @@ enum tfm_plat_err_t nvic_interrupt_enable(void)
 	}
 
 #ifdef MPC_PRESENT
+    mpc_clear_events();
     /* MPC interrupt enabling */
```
tomi-font, line 898: Not a very useful comment by the way. :smile:

tomi-font (approved):

PR #162, opened 2024-07-26
[noup] KMU: Add NOLOAD to linker section for the nRF KMU region

sigvartmh:
This was missed in a previous commit without this the linker will place the initialization of this global variable which is 0 at the address of S_DATA_START in Flash,now it will be discarded.

Ref. NCSDK-25121


Vge0rge (approved):

frkv (approved):
LGTM

LuDuda (approved):

PR #164, opened 2024-08-05
UART fixes for nRF54l

Vge0rge:
Fix some configuration issues with the secure UART which caused build failures or TF-M panics. In essence what these changes do is that:
1) Make sure that if no secure UART is being enabled we don't include any UART configuration code. 
2) Make sure that the configuration of the UART includes all the available UART ports.

### DIFF
--- a/platform/ext/target/nordic_nrf/common/core/target_cfg.c
+++ b/platform/ext/target/nordic_nrf/common/core/target_cfg.c
@@ -872,6 +872,9 @@ enum tfm_plat_err_t init_debug(void)
     return TFM_PLAT_ERR_SUCCESS;
 }
 
+#define NRF_UARTE_INSTANCE(id) NRF_UARTE ## id
+#define NRF_UARTE_INSTANCE_GET(id) NRF_UARTE_INSTANCE(id)
+
 /*----------------- NVIC interrupt target state to NS configuration ----------*/
 enum tfm_plat_err_t nvic_interrupt_target_state_cfg(void)
 {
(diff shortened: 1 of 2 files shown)


vili-nordic (approved):

platform/ext/target/nordic_nrf/common/core/target_cfg.c
```diff
@@ -876,6 +876,16 @@ enum tfm_plat_err_t nvic_interrupt_target_state_cfg(void)
 #elif NRF_SECURE_UART_INSTANCE == 30
     /* UARTE30 is a secure peripheral, so its IRQ has to target S state */
     NVIC_ClearTargetState(NRFX_IRQ_NUMBER_GET(NRF_UARTE30));
+#elif NRF_SECURE_UART_INSTANCE == 00
+    NVIC_ClearTargetState(NRFX_IRQ_NUMBER_GET(NRF_UARTE00));
```
tomi-font, line 880: Couldn't we simply do some macro concatenation here instead? Maybe something like (not sure how that works)
```suggestion
NVIC_ClearTargetState(NRFX_IRQ_NUMBER_GET(NRF_UARTE ##NRF_SECURE_UART_INSTANCE));
```

tomi-font, line 880: Couldn't we simply do some macro concatenation here?

Vge0rge: It needs two macros to do that I think, I did it now, is this what you had in mind?

tomi-font: Yes, lovely!

tomi-font (approved):

tmon-nordic (dismissed):
Commit messages should start with `[nrf noup]`

Vge0rge:
> Commit messages should start with `[nrf noup]`

@tmon-nordic Good catch! I updated both commits.

PR #170, opened 2024-09-24
[nrf noup] platform: nordic_nrf: add support for nrf54l15dk

tomi-font:
And remove support for `nrf54l15pdk` at the same time. Occurences of `NRF54L15_ENGA_XXAA` are replaced by `NRF54L15_XXAA`.

Files from the `nrf54l15dk_nrf54l15_cpuapp` directory that are unused are deleted.
One of them, `tfm_peripherals_config.h`, even had a blatant syntax issue (`#endiff`).


platform/ext/target/nordic_nrf/nrf54l15dk_nrf54l15_cpuapp/ns/cpuarch_ns.cmake
Vge0rge, line 1: Hm, this file and the CMakeLists are required to build the tfm tests AFAIK. Did you manage to build the PSA arch tests and the regression tests with this PR for the new target?

tomi-font: Yes. Those two CMake files manifestly weren't needed.

tomi-font: The `tests` directory inside of this is required, though. It's installed by `sdk-nrf` CMake logic during build. (Not sure if the files in there are actually useful, haven't dug into that.)

Vge0rge (approved):

endre-nordic (approved):

PR #180, opened 2024-11-18
[nrf noup] For Attestation, use message instead of hash

hellesvik-nordic:
Also updated T_COSE library to support this


secure_fw/partitions/initial_attestation/attest_token_encode.c
```diff
@@ -241,6 +241,9 @@ attest_token_encode_start(struct attest_token_encode_ctx *me,
     me->opt_flags  = opt_flags;
     me->key_select = key_select;
 
+    if (opt_flags & TOKEN_OPT_SIGN_MESSAGE) {
```
hellesvik-nordic, line 244: Should I make sure this is only added for nRF54L15, or should we use sign_message for all chips?

lib/ext/t_cose/src/t_cose_sign1_sign.c
```diff
@@ -68,7 +68,7 @@
  * even if key material is not set up or accessible.
  */
 static inline enum t_cose_err_t
-short_circuit_sign(int32_t               cose_algorithm_id,
+    (int32_t               cose_algorithm_id,
```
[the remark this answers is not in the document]
hellesvik-nordic: Yep, that is a mistake. Should be fixed now

lib/ext/t_cose/inc/t_cose_sign1_sign.h
```diff
@@ -103,8 +103,14 @@ struct t_cose_sign1_sign_ctx {
...
 
+/**
+ * By defualt, the cose t_cose_sign1_sign will calculate the hash
+ * of the payload and sign the hash.
+ * Set T_COSE_OPT_SIGN_MESSAGE to instead sign the message(payload)
+ * directly instead.
+ */
 
-
+#define T_COSE_OPT_SIGN_MESSAGE 0x00000010
```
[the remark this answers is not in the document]
hellesvik-nordic: T_COSE_OPT_SHORT_CIRCUIT_SIG exist in both, so I assumed this also has to?
Tried to remove T_COSE_OPT_SIGN_MESSAGE from here, and that worked so no need to have it here I figure

lib/ext/t_cose/tfm_t_cose.cmake
tomi-font, line 1: Why are we making changes to `lib/ext/t_cose` as a noup while this exists upstream?

hellesvik-nordic: Here is the upstream PR: https://review.trustedfirmware.org/c/TF-M/trusted-firmware-m/+/33713.
I will make this PR a draft so I can use it to test the changes for the upstream one.

Vge0rge:
Another strategy was chosen in the end, so this can be closed, right @hellesvik-nordic ?

PR #185, opened 2025-02-04
[nrf fromtree] platform: nrf: ifdef MPC OWNERID configuration

nika-nordic:
Setting OWNERID is not available on all devices.

This PR is needed for nrfx 3.10 integration in NCS

### DIFF
--- a/platform/ext/target/nordic_nrf/common/core/target_cfg.c
+++ b/platform/ext/target/nordic_nrf/common/core/target_cfg.c
@@ -1102,7 +1102,9 @@ static void mpc_configure_override(NRF_MPC_Type *mpc, struct mpc_region_override
 	nrf_mpc_override_endaddr_set(mpc, override->index, override->endaddr);
 	nrf_mpc_override_perm_set(mpc, override->index, override->perm);
 	nrf_mpc_override_permmask_set(mpc, override->index, override->permmask);
+#if defined(NRF_MPC_HAS_OVERRIDE_OWNERID) && NRF_MPC_HAS_OVERRIDE_OWNERID
 	nrf_mpc_override_ownerid_set(mpc, override->index, override->owner_id);
+#endif
 	nrf_mpc_override_config_set(mpc, override->index, &override->config);
 }
 


platform/ext/target/nordic_nrf/common/core/target_cfg.c
```diff
@@ -1102,7 +1102,9 @@ static void mpc_configure_override(NRF_MPC_Type *mpc, struct mpc_region_override
 	nrf_mpc_override_endaddr_set(mpc, override->index, override->endaddr);
 	nrf_mpc_override_perm_set(mpc, override->index, override->perm);
 	nrf_mpc_override_permmask_set(mpc, override->index, override->permmask);
+#if NRF_MPC_HAS_OVERRIDE_OWNERID
```
MarkusLassila, line 1105: I assume this is defined in nrf_mpc.h  in the nrfx 3.10?

MarkusLassila (approved):

tomi-font (changes requested):
This shouldn't be a noup. This code lives in upstream TF-M. Can you submit that upstream and make this a fromtree/fromlist?

nika-nordic:
@tomi-font sure, I guess `upstream` is this repo in TF-M case: https://github.com/TrustedFirmware-M/trusted-firmware-m/blob/main/platform/ext/target/nordic_nrf/common/core/target_cfg.c ? I have also found https://github.com/zephyrproject-rtos/trusted-firmware-m/blob/main/platform/ext/target/nordic_nrf/common/core/target_cfg.c but MPC-related code is not present there

tomi-font:
Correct. Though the GitHub repo is only a read-only mirror, contributions go through Gerrit [here](https://review.trustedfirmware.org/c/TF-M/trusted-firmware-m).
It's interesting/surprising that the Zephyr fork is outdated, I thought it should be more up to date than that.
So actually maybe don't call this a fromlist/fromtree as I think we would do that if we were cherry picking from the Zephyr fork, which we won't (cherry picking directly from upstream TF-M).

nika-nordic:
done: https://review.trustedfirmware.org/c/TF-M/trusted-firmware-m/+/35190

MarkusLassila (approved):
Change the PR name from noup -> fromlist

nika-nordic:
> Change the PR name from noup -> fromlist

done

Vge0rge (approved):

PR #192, opened 2025-04-15
2.1.2 update

tomi-font:
Started from https://github.com/zephyrproject-rtos/trusted-firmware-m/pull/130, re-applied our cherry-picks/noups on top.


degjorva (approved):
range-diffed the no ups and cherry-picks. Except for an additional space in spm_ipc that should be there it looks good

Vge0rge (approved):
Checked the applied patches range-diff and it looks ok to me.

PR #193, opened 2025-04-17
(3.0) [nrf noup] Remove version mismatch CMake warning

tomi-font:
Get rid of the CMake warning which happens when `TFM_VERSION_MANUAL` mismatches `TFM_VERSION_FULL`.
The latter is obtained with Git tags, which cannot be relied on in NCS.


frkv (approved):
LGTM

PR #198, opened 2025-05-26
platform: nordic_nrf: Split the target_cfg for different platforms

Vge0rge:
This splits the target_cfg.c file for the different platfoms. It splits the file into two files one for the nRF54L and one for the nRF91/nRF53 series.

There is no need for the peripherals to be in the same file as the configuration code. There is already a header which provides extern definitions for all the peripherals a device supports which can be used for for source files needing to configure peripherals. So this peripheral definitions are placed in a separate file.

Most of the code is taken unchanged from the target_cfg.c file and the uneeded ifdefs got removed.

The init_debug function got rewritten without changing the logic of it.


Vge0rge:
This will become a fromlist and I will push the PR in the upstream project but I want to get feedback from the CI and the reviews here.

degjorva (approved):
Second look also looks good

Vge0rge:
Closing this, I merged it in upstream TF-M. Cherry picking required more commits and thus I decided to wait until I get it here since this is not urgent.

PR #199, opened 2025-05-27
upmerge 2025.05.27

tomi-font:
Started off from Zephyr's `main`, reapplied our fromtrees/noups on top.

Only difference is the fix to `cmake/version.cmake`; I made a new noup for that which replaces two other ones and properly fixes the TF-M version.

Force pushing this to `main` once ready.


degjorva (approved):
Looks good. Checked with range-diff and only found the expected changes

PR #203, opened 2025-08-05
[nrf fromtree] Platform: nrf54lv10a: Add support for nrf54lv10a

degjorva:
Add initial support for nrf54lv10a
This is based upon the existing nrf54l series of targets Update startup_nrf54l and target_cfg_54l to support this

Change-Id: I0dc11ca4cb39485d29bf595dacc41c3ee6383cdd


platform/ext/target/nordic_nrf/common/nrf54l/config.cmake
```diff
@@ -8,7 +8,7 @@
 
 include(${PLATFORM_PATH}/common/core/config.cmake)
 
-set(SECURE_UART30                       ON         CACHE BOOL      "Enable secure UART" FORCE)
+set(SECURE_UART20                       ON         CACHE BOOL      "Enable secure UART" FORCE)
```
frkv, line 11: Just for my understanding. Changing UART30=>UART20 here is just a default, or is this changed intended for all device types in nRF54L series?

degjorva: Yeah this just changes the default, while we still set the secure uart from dts downstream.

Vge0rge:
Please add comit tags for these, since all of them are merged you need to write [nrf fromtree]

platform/ext/target/nordic_nrf/common/nrf54l/config.cmake
degjorva: Actually not correct. Had to change this upstream to get 54l15 to work

Vge0rge (approved):

frkv (approved):
LGTM (good find removing the FORCE)

PR #208, opened 2025-09-09
[nrf fromtree] platform: nrf54lm20a: Add support for nrf54lm20a

degjorva:
Add initial support for nrf54lm20a
This is based upon the existing nrf54l series of targets Update startup_nrf54l and target_cfg_54l to support this

Change-Id: I6b0c1a818a255263411b67278ab7fcef65063e60


platform/ext/target/nordic_nrf/common/core/CMakeLists.txt
```diff
@@ -310,5 +310,12 @@ install(DIRECTORY   native_drivers
         DESTINATION ${INSTALL_PLATFORM_NS_DIR}/common/core
 )
 
+if(HAS_RRAMC)
+    install(FILES       startup_nrf54l_common.c
+                        startup_nrf54l_common.h
```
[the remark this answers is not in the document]
degjorva: `HAS_RRAMC` is the qualifier we use in TF-m for selecting which startup_files we use and what is included. While we could rewrite that for the L series as it is not the best qualifier, I am going to stick with what we use for everything else here.

platform/ext/target/nordic_nrf/common/core/ns/CMakeLists.txt
```diff
@@ -16,13 +16,38 @@ target_compile_definitions(platform_region_defs
         $<$<BOOL:${NRF_NS_SECONDARY}>:NRF_NS_SECONDARY>
 )
 
+# Determine the startup file based on SOC variant
+if((NRF_SOC_VARIANT MATCHES "^nrf54l1[05]$") OR
+   (TFM_PLATFORM MATCHES "nordic_nrf/nrf54l15dk_nrf54l1[05]_cpuapp"))
+    # nrf54l10 and nrf54l15 share the same startup file
+    set(startup_file "startup_nrf54lx.c")
+elseif((NRF_SOC_VARIANT MATCHES "^nrf54lm20a?$") OR
+       (TFM_PLATFORM MATCHES "nordic_nrf/nrf54lm20dk_nrf54lm20a?_cpuapp"))
```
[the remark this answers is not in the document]
tomi-font: Ah, right! Different regex syntaxes...

tomi-font (commented):
[platform/ext/target/nordic_nrf/common/core/CMakeLists.txt](https://github.com/nrfconnect/sdk-trusted-firmware-m/pull/208#pullrequestreview-3238352124)

platform/ext/target/nordic_nrf/common/core/CMakeLists.txt
tomi-font, line 1: also typo in commit title for the noup

degjorva:
Rebase

tomi-font:
will need rebasing as I just merged https://github.com/nrfconnect/sdk-trusted-firmware-m/pull/213

platform/ext/target/nordic_nrf/common/core/CMakeLists.txt
```diff
@@ -318,5 +318,12 @@ install(DIRECTORY   native_drivers
...
+            DESTINATION ${INSTALL_PLATFORM_NS_DIR}/common/core
+    )
+endif()
```
tomi-font: Shouldn't for example this be upstream? `startup_nrf54l_common.c` lives upstream

degjorva: "When that noup is fixed upstream this should also be added." from the commit message

degjorva: Because these tests only fail due to that revert, without it this builds and runs successfully.

platform/ext/target/nordic_nrf/common/core/ns/CMakeLists.txt
```diff
@@ -16,13 +16,38 @@ target_compile_definitions(platform_region_defs
...
+    # nrf54lm20 has its own startup file
+    set(startup_file "startup_nrf54lm.c")
+elseif((NRF_SOC_VARIANT MATCHES "^nrf54lv10a?$") OR
+       (TFM_PLATFORM MATCHES "nordic_nrf/nrf54lv10dk_nrf54lv10a?_cpuapp"))
+    # nrf54lv10 has its own startup file
+    set(startup_file "startup_nrf54lv.c")
+else()
+    # Default to the original startup file for other targets
+    set(startup_file "startup_${target}.c")
+endif()
```
tomi-font: Also wouldn't it make sense to send this upstream already? Different devices have a different startup file regardless of the revert right?

degjorva: It means I don't know if these need to exist once the noup is reverted. They might, or we might get away with what already exists. As far as I understand how it worked before it seems like we didn't even need the addition of the common file at all. The change that is reverted changed how the build system included the startup files in general so the solution on how these should be included is likely different, both for the ns and the TF-m build, in upstream TF-m. So upstreaming this would just be additional work with zero actual payoff as it likely needs to be changed anyways

PR #243, opened 2026-03-25
[nrf fromtree dirty] Crypto: Allow turning off MBEDTLS_MEMORY_BUFFER_ALLOC_C

frkv:
This removes a faulty requirement to always enable heap regardless if TF-M is using it. 
In the sdk-nrf commit that uses this, we use this to get the build size down for TF-M minimal profile.

The proposed change is already taken in upstream TF-M, changed to [nrf fromtree dirty]

### DIFF
--- a/secure_fw/partitions/crypto/crypto_library.c
+++ b/secure_fw/partitions/crypto/crypto_library.c
@@ -56,8 +56,10 @@ static char mbedtls_version_full[18];
  * \brief Static buffer to be used by Mbed Crypto for memory allocations
  *
  */
+#if defined(MBEDTLS_MEMORY_BUFFER_ALLOC_C)
 #include "config_engine_buf.h"
 static uint8_t mbedtls_mem_buf[CRYPTO_ENGINE_BUF_SIZE] = {0};
+#endif
 
 /* Make sure the library won't print anything through mbedtls_printf */
 static int null_printf(const char *fmt, ...)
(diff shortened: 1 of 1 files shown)


secure_fw/partitions/crypto/crypto_library.c
```diff
@@ -89,12 +91,18 @@ psa_status_t tfm_crypto_core_library_init(void)
...
     mbedtls_memory_buffer_alloc_init(mbedtls_mem_buf,
                                      CRYPTO_ENGINE_BUF_SIZE);
+#endif /* MBEDTLS_MEMORY_BUFFER_ALLOC_C */
 
     mbedtls_platform_set_printf(null_printf);
 
+#if defined(MBEDTLS_MEMORY_BUFFER_ALLOC_C)
     LOG_DBGFMT("[DBG][Crypto] Internal heap size is %d bytes\r\n", sizeof(mbedtls_mem_buf));
+#else
+    LOG_DBG("[DBG][Crypto] No internal heap");
```
SyverHaraldsen: I fail the build with implicit declaration of LOG_DBG
```suggestion
    LOG_DBGFMT("[DBG][Crypto] No internal heap\r\n");
```

simensrostad (approved):

Vge0rge:
The commit message and the PR description is off. Please update the commit tag with the tag "[nrf fromtree dirty] "  so that it is clear that the commit that you took is now merged but you could not apply that cleanly.

You can see this for reference: 
https://github.com/nrfconnect/sdk-trusted-firmware-m/commit/c1bb2ac7d407d26f0c40dbc52d5e5f168c2a698e

Vge0rge (approved):

tomi-font (changes requested):
The commit hash is invalid, please re-cherry-pick the merged commit from upstream

tomi-font (approved):

PR #244, opened 2026-03-26
[nrf noup] crypto: Fix missing WPA3 prerequisite for PAKE

AntonZma:
Fixes the build time error stating that PAKE is enabled, but not all prerequisites if WPA3 is the only activated PAKE algorithm.


tomi-font (approved):
`squash!` => `nrf-squash!`

tomi-font (approved):

degjorva (approved):

PR #245, opened 2026-04-07
[nrf noup]platform: nordic_nrf: Add support for TFM_SHARED_INSTANCE

degjorva:
UART instances would be locked permanently even when TFM_SHARED_INSTANCE was set.
Update to allow shared instance for UART on
54L and 71 series devices.
TFM_SHARED_INSTANCE is a NCS config
which is why this is a noup.

### DIFF
--- a/platform/ext/target/nordic_nrf/common/core/native_drivers/spu.c
+++ b/platform/ext/target/nordic_nrf/common/core/native_drivers/spu.c
@@ -351,7 +351,9 @@ void spu_peripheral_config_secure(const uint32_t periph_base_address, bool perip
 
 	nrf_spu_periph_perm_secattr_set(nrf_spu, index, true /* Secure */);
 	nrf_spu_periph_perm_dmasec_set(nrf_spu, index, true /* Secure */);
-	nrf_spu_periph_perm_lock_enable(nrf_spu, index);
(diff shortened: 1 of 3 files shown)


tomi-font (commented):
Any chance we could upstream this instead of building on top of noups?
By the way, there is already one instance of `CONFIG_TFM_LOG_SHARE_UART` in upstream TF-M (probably by mistake, but anyway):
```
platform/ext/target/nordic_nrf/common/core/cmsis_drivers/Driver_USART.c:#if !(DOMAIN_NS == 1U) && defined(CONFIG_TFM_LOG_SHARE_UART) && (defined(NRF_SPU) || defined(NRF_SPU00))
```

degjorva:
@tomi-font Sure, but that is a bigger task, this is a bug fix because it is not correctly implemented on 54L. The instance you pointed out should probably be removed until we do have some kind of upstream solution for it.

tomi-font:
Fair point, can you just point to the noup this is amending? possibly with a simple `nrf-squash! <commit title>`

degjorva:
Sure, done

tomi-font (approved):

endre-nordic (approved):

Vge0rge (approved):

PR #248, opened 2026-04-20
[nrf fromtree]Platform: nordic_nrf: select TFM_PERIPHERAL_STD_UART fr…

degjorva:
…om NRF_SECURE_UART_INSTANCE

TFM_PERIPHERAL_STD_UART was hardcoded to TFM_PERIPHERAL_UARTE30, which breaks nrf54lv10a (NRF_SECURE_UART_INSTANCE=20) since tfm_peripheral_uarte30 is not compiled in, causing an undefined reference at link time. Select the right peripheral from NRF_SECURE_UART_INSTANCE, matching how target_cfg_54l.c and target_cfg.h already use it.

Change-Id: I5aa90808a0592a6da3ecef6073ad06814ab609d5

(cherry picked from commit d404a04fb5027775a5bf8e7a0d09368105d2f942)


tomi-font (approved):
missing space after `]`

Vge0rge (approved):

nordic-piks (approved):

PR #250, opened 2026-04-29
[nrf fromtree] secure_fw: Add missing return value for psa_map_outvec

simonduq:
This fixes:
warning: control reaches end of non-void function [-Wreturn-type].

### DIFF
--- a/secure_fw/partitions/lib/runtime/psa_api_ipc.c
+++ b/secure_fw/partitions/lib/runtime/psa_api_ipc.c
@@ -149,7 +149,7 @@ void psa_unmap_invec(psa_handle_t msg_handle, uint32_t invec_idx)
 
 void *psa_map_outvec(psa_handle_t msg_handle, uint32_t outvec_idx)
 {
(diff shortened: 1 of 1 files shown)


tomi-font (changes requested):
Thanks for the fix, but we don't do noups for this.
This has already been fixed upstream, so instead fromtree this commit: https://github.com/TrustedFirmware-M/trusted-firmware-m/commit/2e1e202b5a4406746c7a06af49113976d4fb3d44

simonduq:
Ah excellent, thank you, I was not sure indeed if noup was correct here. And missed that there was a fix upstream.

simonduq:
Done.

tomi-font:
You still need to update the commit itself. Cherry-pick the upstream commit with `-x` and add `[nrf fromtree]`.

simonduq:
Oh, ok, should be good now.

tomi-font (approved):
Perfect!

degjorva (approved):

PR #253, opened 2026-05-08
TF-M changes for Mbed TLS update to v4.1.0

tomi-font:
manifest-pr-skip


Vge0rge:
Now tfm_crypto_library_get_info returns an un-initialized but since it is static thats not a big issue and it will get updated with the new TF-M anyway. Looks good otherwise.

Vge0rge (approved):

PR #255, opened 2026-05-22
Update to TF-M 2.3

Vge0rge:
* da23da453 [nrf noup] remove usage of mbedtls/version.h -- Skipped, not needed 
* a797b52fb [nrf noup] remove references to psa_open_key(), psa_close_key(), and psa_key_handle_t -- Skipped, not needed
* fa1b039cf [nrf noup] platform: nordic_nrf: Fix bug in nrfx errata -- Skipped,  it was reverted afterwards
* 87015607e [nrf noup] attestation: Prevent errors of multiple defs of ARRAY_LENGTH -- Skipped, it seems that now they don't define ARRAY_LENTH :) 
* cdcdebf94 [nrf noup] cmake: tools: Dont add a custom command for running the manifest -- Skipped, the add custom does not exist anymore
* 01140ced6 [nrf noup] Remove check for p256-m -- Replaced by nrf noup] crypto: Remove TF-PSA-Crypto internal targets


cmake/remote_library.cmake
```diff
@@ -134,6 +134,11 @@ function(fetch_remote_library)
     else()
         string(TOUPPER ${ARG_LIB_NAME} ARG_LIB_NAME_UPCASE)
         set(FETCHCONTENT_SOURCE_DIR_${ARG_LIB_NAME_UPCASE} ${${ARG_LIB_SOURCE_PATH_VAR}})
+
+        # User-supplied source path: do not let MakeAvailable add_subdirectory it,
```
tomi-font, line 138: Commit message typo (missing word): `because FetchContent_MakeAvailable will always when`

cmake/install.cmake
```diff
@@ -105,9 +105,6 @@ if (TFM_PARTITION_CRYPTO)
                             ${INTERFACE_INC_DIR}/psa/crypto.h
                             ${INTERFACE_INC_DIR}/psa/crypto_compat.h
                             ${INTERFACE_INC_DIR}/psa/crypto_driver_common.h
-                            ${INTERFACE_INC_DIR}/psa/crypto_driver_contexts_composites.h
-                            ${INTERFACE_INC_DIR}/psa/crypto_driver_contexts_key_derivation.h
-                            ${INTERFACE_INC_DIR}/psa/crypto_driver_contexts_primitives.h
```
tomi-font, line 110: Shouldn't this be squashed into `[nrf noup] inteface: Remove headers existing in the Oberon PSA core`?

platform/ext/target/nordic_nrf/nrf54lv10dk_nrf54lv10a_cpuapp/CMakeLists.txt
```diff
@@ -50,6 +50,7 @@ endif()
 
 install(FILES       RTE_Device.h
                     device_cfg.h
+                    ns/CMakeLists.txt
```
tomi-font: I think you can drop this commit (`[nrf noup] Revert "platform: nordic_nrf: Fix build failure for nRF54L… `). The macro renaming already happened as part of 015f7b7f871334a1cb30dc629bf0ab59abb64508 which you have in your branch, which is why this revert is now doing only part of what it should.

platform/ns/toolchain_ns_GNUARM.cmake
```diff
@@ -79,7 +79,7 @@ add_compile_options(
     -Wno-format
     -Warray-parameter
     -Wno-unused-but-set-variable
-    -Wnull-dereference
+     #-Wnull-dereference
```
tomi-font, line 82: dereferece => dereference

also would be good to mention 4fdf89b5410aa0ebb5d682d664524ae709729a2d in your commit message

secure_fw/spm/core/arch/tfm_arch_v8m_main.c
```diff
@@ -23,6 +23,21 @@
 
 #include "compiler_ext_defs.h" /* Keep last. */
 
+#if defined(__ARM_ARCH_8M_MAIN__)
+/** \brief SCnSCB Coprocessor Power Control Register Definitions */
+#define SCnSCB_CPPWR_SUS11_Pos             23U                                         /*!< CPPWR: SUS11 Position */
+#define SCnSCB_CPPWR_SUS11_Msk             (1UL << SCnSCB_CPPWR_SUS11_Pos)             /*!< CPPWR: SUS11 Mask */
+
+#define SCnSCB_CPPWR_SU11_Pos              22U                                         /*!< CPPWR: SU11 Position */
```
tomi-font, line 31: In the message of this commit (`[nrf noup] Add missing CPPWR for ARM V8M`) mention the path to the patch file in TF-M? could be a way to see whether it's still needed

cmake/remote_library.cmake
Vge0rge: I rewrote it, please check again.

platform/ext/target/nordic_nrf/nrf54lv10dk_nrf54lv10a_cpuapp/CMakeLists.txt
Vge0rge: You are right, I dropped the commit.

secure_fw/spm/core/arch/tfm_arch_v8m_main.c
Vge0rge: I added the filename in the commit message.

lib/ext/qcbor/CMakeLists.txt
```diff
@@ -33,6 +33,8 @@ fetch_remote_library(
         GIT_PROGRESS        TRUE
 )
 
+add_subdirectory(${QCBOR_PATH} ${CMAKE_CURRENT_BINARY_DIR}/qcbor)
```
tomi-font, line 36: Combine this (`[nrf noup] lib: qcbor: Add target`) with `[nrf noup] cmake: remote_library: Skip calling add_subdirectory`? Plus could we just check the library name in the `fetch_remote_library` function instead so that all the changes would be in the same place?

Vge0rge: I would argue that it is better as is. Because I want to avoid calling add_subdirectory to all the other projects that use the same mechanism. So I prefer it like that because if they add a new project later it will fail to build, which I consider an advantage because we will notice it.  About the commits it still reads a bit better to me to have them separate but I don't have a strong opinion.

tomi-font: I agree, what I meant was to do something like that in `cmake/remote_library.cmake`:
```
if(NOT ARG_LIB_NAME_UPCASE STREQUAL QCBOR)
    list(APPEND ARG_FETCH_CONTENT_ARGS
     SOURCE_SUBDIR _fetch_remote_library_no_add_subdirectory)
endif()
```

Vge0rge: This will probably work but I don't really fancy it to be honest. Since we already have to do kind of a hack it is better to be consistent in this and not have target specific details in a generic function.

tomi-font: Indeed the whole thing is a hack but at least it would all be in one place which to me seems like a better idea. anyway as you want

tomi-font (approved):

degjorva (approved):
Nothing missing. NOUPs make sense.

tomi-font (commented):
Minor thing, but I would force-push the `zephyr_tf-m_v2.3.0` branch to `18374359a845da32700fde30e8d76a737df59f84` which contains both zep noups you just added plus the 7120 MRAMC service fromtree.

PR #263, opened 2026-07-22
[nrf fromtree] platform: nordic_nrf: apply nRF7120 RAM retention directly

erdemsimsek:
This PR brings the upstream changes to the `sdk-trusted-firmware-m`
Changes can be seen at https://review.trustedfirmware.org/c/TF-M/trusted-firmware-m/+/52584

Scope of the changes basically
- Previously ram-retention for system-off was clear at the `system-off` phase and then reenabled for the regions defined in the device tree file. Instead of that, ram-retention cleared at the beginning of the application. This allow developers to have configure ram-retention with run-time APIs.
- Previously system-off ram retention requests were cached and applied later. Instead of that, applied directly.

manifest-pr-skip

### DIFF
--- a/platform/ext/target/nordic_nrf/common/core/CMakeLists.txt
+++ b/platform/ext/target/nordic_nrf/common/core/CMakeLists.txt
@@ -184,12 +184,17 @@ if(SOC_NRF7120_TFM_MRAMC_SERVICE)
     )
 endif()
 
-# RAM-control service (set by zephyr's -DNRF_RAM_CTRL_SERVICE=ON).
-if(NRF_RAM_CTRL_SERVICE)
+# nRF7120 disables RAM retention during secure startup. Other platforms need
+# the helper only when the RAM-control service is enabled.
+if(NRF_RAM_CTRL_SERVICE OR target STREQUAL "nrf7120")
     target_sources(platform_s
         PRIVATE
             ${HAL_NORDIC_PATH}/nrfx/helpers/nrfx_ram_ctrl.c
     )
+endif()
+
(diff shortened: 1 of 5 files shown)


tomi-font (changes requested):
`git cherry-pick -x`

tomi-font:
\+ don't forget `manifest-pr-skip`

erdemsimsek:
> + don't forget `manifest-pr-skip`

I couldn't find the `manifest-pr-skip` label. Is it in the manifest PR on this PR?

tomi-font:
> I couldn't find the manifest-pr-skip label. Is it in the manifest PR on this PR?

no, it's a line you add to the PR description

tomi-font (approved):

travis3630 (approved):

adrianiainlam:
Hi, what is the status of this PR? Do we have an ETA? This is blocking #272 which in turn blocks https://github.com/nrfconnect/sdk-nrf/pull/30564

tomi-font (approved):

rob-robinson-14:
This PR is currently blocked by: 

https://github.com/zephyrproject-rtos/trusted-firmware-m/pull/220
https://github.com/zephyrproject-rtos/zephyr/pull/115539

and then integration into sdk-zehpyr / sdk-nrf.


I can try and break up the sdk-nrf PR in two (tfm / other) So that this PR is not blocked for as long. https://github.com/nrfconnect/sdk-nrf/pull/30587 but there may be some definition errors so not 100% sure this will be possible.

tomi-font:
@sigvartmh in the end it's probably simplest to just make a dirty cherry-pick

rob-robinson-14 (approved):

PR #277, opened 2026-08-21
[ncs-v3.4-branch] update to TF-M v2.3.1

tomi-font:
Started from upstream `TF-Mv2.3.1` tag, reapplied all the `zep` and `nrf` patches on [`ncs-v3.4-branch`](https://github.com/nrfconnect/sdk-trusted-firmware-m/commits/ncs-v3.4-branch/) (with some adjustments as needed).

manifest-pr-skip


frkv (approved):
LGTM (TF-M commits check out)

Vge0rge:
Is there a reason that you kept the deleting of the cmake install lines here: https://github.com/nrfconnect/sdk-trusted-firmware-m/pull/277/changes/1032bd6242a6c786b1f42604a44a8101b0609a99 

If not I would say just make the commit identical with the PR that goes to main.

tomi-font:
> Is there a reason that you kept the deleting of the cmake install lines here: [1032bd6](https://github.com/nrfconnect/sdk-trusted-firmware-m/commit/1032bd6242a6c786b1f42604a44a8101b0609a99)
> 
> If not I would say just make the commit identical with the PR that goes to main.

I just did the minimal changes for the update to this branch. But in the v3.4 branch we also have `-DTFM_INSTALL_TF_PSA_CRYPTO_HEADERS=OFF` so I'll drop the part you mention.

Vge0rge (approved):

